PR #2, opened 2021-07-23
Add extra guards to complex type handlers

mmv:
Complex types for CefValue such as Binary, Dictionary and List can take
a different owner (eg: a Dictionary may be assigned to an element of a
List), which turns the handle invalid. Ideally we should discard the
handles for such elements as quickly as possible, but in practice the
wrapper objects in .NET can stay in memory for a long time before the GC
decides to run their finalizers. By running the finalizers late, the
probability that the handle has become invalid is high.

In a previous commit, we wrapped most usages of complex values within
`using` blocks, calling the Dispose immediately instead of the finalizer
a lot later, which reduced incidence of crashes. Apparently that was not
enough, so with this commit we add a few more protection on remaining
usages of Lists, Dictionaries and Binaries.

### DIFF
--- a/CefGlue.BrowserProcess/Serialization/V8ValueSerialization.cs
+++ b/CefGlue.BrowserProcess/Serialization/V8ValueSerialization.cs
@@ -109,7 +109,12 @@ namespace Xilium.CefGlue.BrowserProcess.Serialization
             switch (cefValue.GetValueType())
             {
                 case CefValueType.Binary:
-                    var nativeValue = CefValueSerialization.FromCefBinary(cefValue.GetBinary(), out var kind);
+                    object nativeValue;
+                    CefValueSerialization.BinaryMagicBytes kind;
+                    using (var binaryValue = cefValue.GetBinary())
+                    {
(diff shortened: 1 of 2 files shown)


CefGlue.BrowserProcess/Serialization/V8ValueSerialization.cs
```diff
@@ -109,7 +109,12 @@ public static CefV8Value SerializeCefValue(CefValueWrapper cefValue)
             switch (cefValue.GetValueType())
             {
                 case CefValueType.Binary:
-                    var nativeValue = CefValueSerialization.FromCefBinary(cefValue.GetBinary(), out var kind);
+                    object nativeValue;
+                    CefValueSerialization.BinaryMagicBytes kind;
+                    using (var binaryValue = cefValue.GetBinary())
+                    {
+                        nativeValue = CefValueSerialization.FromCefBinary(cefValue.GetBinary(), out kind);
```
joaompneves, line 116: @mmv 
we already talked about this... so that it does not get forgotten, i'll leave here the note:
cefValue.GetBinary() -> binaryValue

PR #12, opened 2021-08-13
memleak: break object reference cycle so GC works

mmv:
The GC is smart enough to deal with reference cycles. But in this
particular case the GC can't work its magic, because the referential
cycle has a part in the native land, to which the GC has no visibility.

This cycle works with something like:

  CefBrowser -> [ native ] -> CefClient -> (...) -> CefBrowser

CefClient is kept active because there's a native reference to it, but
the GC doesn't know what a native reference is... all it knows is that
the object was pinned (we must pin it to keep the native reference
valid). So the GC only sees something like

  (pinned root) -> CefClient -> (...) -> CefBrowser

This prevents the GC from collecting CefBrowser, which prevents the
native reference from being cleared, which prevents CefClient to be
cleared.

We break this cycle by making objects along the chain (represented above
by "(...)") clear their reference to CefBrowser upon disposal.

### DIFF
--- a/CefGlue.Common/ObjectBinding/NativeObjectRegistry.cs
+++ b/CefGlue.Common/ObjectBinding/NativeObjectRegistry.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using Xilium.CefGlue.Common.Events;
@@ -5,7 +6,7 @@ using Xilium.CefGlue.Common.Shared.RendererProcessCommunication;
 
 namespace Xilium.CefGlue.Common.ObjectBinding
 {
-    internal class NativeObjectRegistry
+    internal class NativeObjectRegistry : IDisposable
     {
         private CefBrowser _browser;
         private readonly Dictionary<string, NativeObject> _registeredObjects = new Dictionary<string, NativeObject>();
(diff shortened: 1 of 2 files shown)


CefGlue.Common/ObjectBinding/NativeObjectRegistry.cs
```diff
@@ -97,5 +98,10 @@ private void SendRegistrationMessage(NativeObject obj)
             // TODO target main frame?
             _browser.GetMainFrame().SendProcessMessage(CefProcessId.Browser, cefMessage);
         }
+
+        public void Dispose()
+        {
+            _browser = null;
```
joaompneves, line 104: what if we use a weak ref instead? isnt it better?
we have more classes that store the browser... i got your explanation but isnt there more classes where the pattern exists? should we apply the same to those?

tbragaf (approved):

CefGlue.Common/ObjectBinding/NativeObjectRegistry.cs
mmv: That's a good idea — using a WeakRef for handler types that ought to be held by the reference counter anyway might be an interesting strategy towards avoiding these "cross domain" reference loops. But since that will require a bit more analysis I'd prefer to ship this fix right now and do further improvements in a separate task/PR.

joaompneves (approved):

PR #13, opened 2021-08-20
attempted fix to pipeline versioning

mmv:


Directory.Build.props
```diff
@@ -9,7 +9,10 @@
 
     <PropertyGroup>
         <PackageOutputPath>$(MSBuildProjectDirectory)\..\Nuget\output</PackageOutputPath>
-        <Version>91.4472.9</Version>
+        <VersionPatchNumber Condition="'$(Build_BuildId)' != ''">$(Build_BuildId)</VersionPatchNumber>
```
joaompneves, line 12: is this coming from azure? we had this in the past and the problem is that build ids can bee too big (overflowing)

PR #16, opened 2021-09-02
LoadCefGlueBrowserProcess file from the currently executing assembly

adirh3:
Haven't seen any word regarding to contributing, so I hope this is accepted.

CefGlue is trying to load dependencies from `AppDomain.CurrentDomain.BaseDirectory` which works for most scenarios, but in case that the DLL that uses CefGlue is in another directory it won't work.
For example, in my case, I have an app that loads user plugins from a different directory, causing this to fail.

Let me know if any changes are required.

### DIFF
--- a/CefGlue.Common/CefRuntimeLoader.cs
+++ b/CefGlue.Common/CefRuntimeLoader.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Reflection;
 using Xilium.CefGlue.Common.Handlers;
 using Xilium.CefGlue.Common.Shared;
 
@@ -27,18 +28,14 @@ namespace Xilium.CefGlue.Common
             settings.UncaughtExceptionStackSize = 100; // for uncaught exception event work properly
 
             var path = AppDomain.CurrentDomain.BaseDirectory;
-
-            var subprocessPath = Path.Combine(path, BrowserProcessFileName);
-            if (!File.Exists(subprocessPath))
+            if (!TryGetSubprocessPath(path, out string subprocessPath))
             {
-                subprocessPath = Path.Combine(path, "CefGlueBrowserProcess", BrowserProcessFileName);
-                
-                if (!File.Exists(subprocessPath))
-                {
+                // The executing DLL might not be in the current domain directory (plugins scenario)
+                path = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+                if(!TryGetSubprocessPath(path, out subprocessPath))
                     throw new FileNotFoundException($"Unable to find \"{subprocessPath}\"");
-                }
             }
-            
+
             settings.BrowserSubprocessPath = subprocessPath;
 
             switch (CefRuntime.Platform)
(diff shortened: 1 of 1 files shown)


adirh3:
I read, understood and accept the OutSystems CLA

CefGlue.Common/CefRuntimeLoader.cs
```diff
@@ -77,6 +76,22 @@ private static void InternalInitialize(CefSettings settings = null, KeyValuePair
...
+                subprocessPath = Path.Combine(baseDirectory, "CefGlueBrowserProcess", BrowserProcessFileName);
+
+                if (!File.Exists(subprocessPath))
+                {
+                    return false;
+                }
+            }
+
+            return true;
+        }
```
joaompneves, line 93:
```suggestion
        private static IEnumerable<string> GetSubProcessPaths()
        {
            var baseDirectory = AppDomain.CurrentDomain.BaseDirectory;
            yield return Path.Combine(baseDirectory, BrowserProcessFileName);
            yield return Path.Combine(baseDirectory, CefGlueBrowserProcessDirName, BrowserProcessFileName);
        
            // The executing DLL might not be in the current domain directory (plugins scenario)
            baseDirectory = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
            yield return Path.Combine(baseDirectory, BrowserProcessFileName);
            yield return Path.Combine(baseDirectory, CefGlueBrowserProcessDirName, BrowserProcessFileName);
        }
```

CefGlue.Common/CefRuntimeLoader.cs
```diff
@@ -27,18 +28,16 @@ private static void InternalInitialize(CefSettings settings = null, KeyValuePair
             settings.UncaughtExceptionStackSize = 100; // for uncaught exception event work properly
 
             var path = AppDomain.CurrentDomain.BaseDirectory;
-
-            var subprocessPath = Path.Combine(path, BrowserProcessFileName);
-            if (!File.Exists(subprocessPath))
+            if (!TryGetSubprocessPath(path, out string subprocessPath))
```
joaompneves, line 31:
```suggestion
            var subProcessPaths = GetSubProcessPaths().ToList();
            var subProcessPath = subProcessPaths.FirstOrDefault(path => File.Exists(path));
            if (string.IsNullOrWhiteSpace(path))
            {
                throw new FileNotFoundException($"Unable to find SubProcess. Probed locations: {string.Join(", ", "\" + subProcessPaths + \"}");
            }
```

joaompneves (changes requested):
Thank you for your contribution.
I made a suggestion to your changes to ease debugging (and understanding) in case of failure.

CefGlue.Common/CefRuntimeLoader.cs
adirh3: @joaompneves committed, looks cleaner, thanks!

joaompneves: Its not building, please build and test it.

joaompneves:
Build failing

PR #27, opened 2021-12-09
RMAC-8338: Force update window boundaries

cangosta:
@joaompneves added logs to the NSView and for the NativeControlHost. All the values were correct. Everything indicates that the problem is on CEF side.

### DIFF
--- a/CefGlue.Avalonia/Platform/ExtendedAvaloniaNativeControlHost.cs
+++ b/CefGlue.Avalonia/Platform/ExtendedAvaloniaNativeControlHost.cs
@@ -1,13 +1,8 @@
 using System;
 using Avalonia;
-// using System.Collections.Generic;
-// using System.Diagnostics;
 using Avalonia.Controls;
-// using Avalonia.Controls.Platform;
 using Avalonia.Platform;
 using Avalonia.Threading;
-
-// using Avalonia.VisualTree;
 
 namespace Xilium.CefGlue.Avalonia.Platform
 {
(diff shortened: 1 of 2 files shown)


CefGlue.Avalonia/Platform/ExtendedAvaloniaNativeControlHost.cs
```diff
@@ -22,5 +39,44 @@ protected override void DestroyNativeControlCore(IPlatformHandle control)
...
+            if (!Bounds.IsEmpty && _isAttached)
+            {
+                // try delay native host position update, because running right away seems to have no effect sometimes
+                DispatcherTimer.RunOnce(() =>
+                {
+                    if (_isAttached)
+                    {
+                        TryUpdateNativeControlPosition();
+                    }
+                }, TimeSpan.FromMilliseconds(500));
```
cangosta, line 78: @joaompneves I had to increase this time. having a small timepan here did not solve the problem when the machine is slow. IMO,  increasing the time is not a problem since the only practical effect is that the user may notice the update in the internal view (since the window size is correct from the beginning).

joaompneves: obviously its better than the problem it self ... although its meh and if you have a machine even slower the problem can still occur... have you checked other alternatives described here: https://magpcss.org/ceforum/viewtopic.php?f=6&t=16341

cangosta: I'll check that

brunocunhasilva (approved):

PR #29, opened 2021-12-30
RDOIAT-369 Allow async native methods execution

joaompneves:
Currently when we specify a value greater than 1 to the MaxNativeMethodsParallelCalls property, there's a risk that the calls order might not be preserved. This PR changes the current execution behavior as follows:
- sync methods wont run in parallel (they always run sequentially, one after the other)
- async methods (methods that return a Task) can run in parallel... after the execution is yield. This allows the target method execution to take some measures in order to honor the call order. 

The MaxNativeMethodsParallelCalls property was removed and it's now the responsibility of the consumers to control this behavior (when needed).


cangosta (approved):

Xilium.CefGlue.sln
```diff
@@ -23,21 +23,14 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
...
+		Directory.Build.targets = Directory.Build.targets
 	EndProjectSection
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "CefGlue.Demo.WPF", "CefGlue.Demo.WPF\CefGlue.Demo.WPF.csproj", "{8E758128-CA79-4533-963E-91A1BF0558FB}"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "CefGlue.BrowserProcess", "CefGlue.BrowserProcess\CefGlue.BrowserProcess.csproj", "{1E60FB67-049E-4877-92A9-D5480F81A324}"
 EndProject
-Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "NuGet", "NuGet", "{953C06E5-DABF-4958-8640-62147CD1DFAD}"
-	ProjectSection(SolutionItems) = preProject
-		NuGet\CefGlue.Avalonia.nuspec = NuGet\CefGlue.Avalonia.nuspec
```
alvesmiguel1, line 34: Are these files still being used?

joaompneves: They were removed a while ago ... but were still in the sln

CefGlue.Tests/Javascript/NativeObjectInteropTests.cs
```diff
@@ -129,5 +157,82 @@ public async Task NativeObjectMethodResultIsReturned()
 
             Assert.AreEqual(nativeObject.MethodWithReturn(), result[0]);
         }
+        
+        [Test]
+        public void AsyncMethodsCanExecuteSimultaneously()
```
alvesmiguel1, line 162: Don't know if we already have it, but can we have a test that verifies the method execution order of more than one method executed synchronously?

alvesmiguel1: Also, would a test executing both a sync and an async method have some value? (If I understood correctly, the sync method should be executed first)

CefGlue.Common/ObjectBinding/NativeObject.cs
```diff
@@ -1,33 +1,72 @@
...
+            });
+        }
 
-        public MethodInfo GetNativeMethod(string javascriptMethodName)
+        protected static IDictionary<string, NativeMethod> GetObjectMembers(object obj)
         {
-            _methods.TryGetValue(javascriptMethodName, out var nativeMethod);
-            return nativeMethod;
+            var methods = obj.GetType().GetMethods(BindingFlags.Instance | BindingFlags.Public).Where(m => !m.IsSpecialName);
+            return methods.Select(m => new { Member = m, JavascriptName = m.Name.Substring(0, 1).ToLowerInvariant() + m.Name.Substring(1) })
```
alvesmiguel1, line 68: I think you can have just the ToDictionary statement right away

CefGlue.Tests/Javascript/NativeObjectInteropTests.cs
joaompneves: - regarding the execution order this test tries to ensure that but its not feasible to create a test that truly guarantees the order because the original problem was random and depended on the cpu scheduling
- methods are executed by their call order ... independently if they are sync or async (didn't understood your suggestion quite well)

alvesmiguel1 (dismissed):

CefGlue.Tests/Javascript/NativeObjectInteropTests.cs
alvesmiguel1: The second suggestion was more like a brainstorming, but once I finished the code review I understand that it does not make sense

alvesmiguel1 (dismissed):

cangosta (dismissed):

CefGlue.Tests/Javascript/NativeObjectInstrospectionTests.cs
```diff
@@ -31,25 +35,26 @@ private void PrivateMethod()
...
-            var members = NativeObjectAnalyser.AnalyseObjectMembers(new CustomObject());
+            var members = new HashSet<string>(new NativeObject("", new CustomObject()).MethodsNames);
 
-            Assert.AreEqual(5, members.Count); // object members + 1 public method
-            Assert.IsTrue(members.ContainsKey("methodWithParams"));
-            Assert.IsTrue(members.ContainsKey("toString"));
-            Assert.IsTrue(members.ContainsKey("getHashCode"));
-            Assert.IsTrue(members.ContainsKey("getType"));
-            Assert.IsTrue(members.ContainsKey("equals"));
+            Assert.AreEqual(6, members.Count); // object members + 2 public methods
```
cangosta, line 51: @joaompneves shouldn't the AsyncGenericMethod be one of the members?

PR #31, opened 2022-01-05
RDOIAT-374: Improve objects serialization performance

joaompneves:
Revamp objects serialization between render process and browser process.
We now use json to pass information in the following cases:
- strings
- datetimes
- binaries
- lists
- complex objects (like dictionaries)

The information is serialized on the render process and passed to the browser process as json in the aforementioned situations. It is then the browser responsible for parsing the json into JS objects. 
This way we are able to save an extra serialization from cef value into a v8 value.
On my tests this could make this transfer almost 10x faster.


cangosta (approved):

cangosta (commented):
There are some tests failling

joaompneves:
> There are some tests failling

I'm aware ... I need to rewrite the tests add add new tests also

alvesmiguel1 (dismissed):

cangosta (dismissed):

cangosta (approved):

PR #49, opened 2022-05-06
Unsubscribe to events on close and cleanup to release the references

brunocunhasilva:
Solve leaks related with a few events that need to be unsubscribed when closing the browser
image

Also release the references in the _registeredObjects dictionary:
image

### DIFF
--- a/CefGlue.Common.Shared/RendererProcessCommunication/PipeServer.cs
+++ b/CefGlue.Common.Shared/RendererProcessCommunication/PipeServer.cs
@@ -10,7 +10,7 @@ namespace Xilium.CefGlue.Common.Shared.RendererProcessCommunication
     {
         private const int MaxErrorsAllowed = 5;
 
-        private readonly CancellationTokenSource _cancellationTokenSource = new CancellationTokenSource();
+        private CancellationTokenSource _cancellationTokenSource = new CancellationTokenSource();
 
         public event Action<string> MessageReceived;
 
@@ -38,12 +38,18 @@ namespace Xilium.CefGlue.Common.Shared.RendererProcessCommunication
                         }
                     }
                 }
+
+                var cancellationTokenSource = _cancellationTokenSource;
+                _cancellationTokenSource = null;
+                cancellationTokenSource.Dispose();
             });
         }
 
         public void Dispose()
         {
-            _cancellationTokenSource.Cancel();
+            _cancellationTokenSource?.Cancel();
+            // release the MessageReceived handlers to prevent any possible memory leak
+            MessageReceived = null;
         }
 
         private void HandleClientConnected(Stream pipe)
--- a/CefGlue.Common/CommonBrowserAdapter.cs
+++ b/CefGlue.Common/CommonBrowserAdapter.cs
@@ -451,6 +451,8 @@ namespace Xilium.CefGlue.Common
                 return false; 
             }
 
+            Control.GotFocus -= HandleGotFocus;
+            Control.SizeChanged -= HandleControlSizeChanged;
             Control.DestroyRender();
             Cleanup(browser);
 
@@ -459,6 +461,11 @@ namespace Xilium.CefGlue.Common
 
         protected void Cleanup(CefBrowser browser)
         {
+            if (_crashServerPipe != null)
+            {
+                _crashServerPipe.MessageReceived -= OnChildProcessCrashed;    
+            }
+            
             _crashServerPipe?.Dispose();
 
             browser.Dispose();
--- a/CefGlue.Common/CommonBrowserAdapter.cs
+++ b/CefGlue.Common/CommonBrowserAdapter.cs
@@ -451,8 +451,6 @@ namespace Xilium.CefGlue.Common
                 return false; 
             }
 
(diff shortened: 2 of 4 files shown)


CefGlue.Common/CommonBrowserAdapter.cs
```diff
@@ -459,6 +461,11 @@ protected virtual bool OnBrowserClose(CefBrowser browser)
 
         protected void Cleanup(CefBrowser browser)
         {
+            if (_crashServerPipe != null)
+            {
+                _crashServerPipe.MessageReceived -= OnChildProcessCrashed;
```
joaompneves, line 466: looking at the object graph this is being held because the WaitForConnectionAsync (https://github.com/OutSystems/CefGlue/blob/4472/CefGlue.Common.Shared/RendererProcessCommunication/PipeServer.cs#L28) is still running. Once the  _crashServerPipe gets disposed that should go away. Are you sure that you waited long enough for this to be freed?

In case that there's really a problem, clear all the handlers inside  the pipe server instead.

CefGlue.Common/CommonBrowserAdapter.cs
```diff
@@ -459,6 +461,11 @@ protected virtual bool OnBrowserClose(CefBrowser browser)
 
         protected void Cleanup(CefBrowser browser)
         {
+            if (_crashServerPipe != null)
+            {
+                _crashServerPipe.MessageReceived -= OnChildProcessCrashed;    
+            }
```
joaompneves, line 467: Also noticed that there's a missing dispose of 
https://github.com/OutSystems/CefGlue/blob/4472/CefGlue.Common.Shared/RendererProcessCommunication/PipeServer.cs#L13

here:
https://github.com/OutSystems/CefGlue/blob/4472/CefGlue.Common.Shared/RendererProcessCommunication/PipeServer.cs#L40

CefGlue.Common/CommonBrowserAdapter.cs
```diff
@@ -451,6 +451,8 @@ protected virtual bool OnBrowserClose(CefBrowser browser)
                 return false; 
             }
 
+            Control.GotFocus -= HandleGotFocus;
+            Control.SizeChanged -= HandleControlSizeChanged;
```
joaompneves, line 455: whats this for? any real evidence of problems here?... Control is an internal object ... all the references should be kept locally

joaompneves (changes requested):

CefGlue.Common/CommonBrowserAdapter.cs
brunocunhasilva: "Once the _crashServerPipe gets disposed that should go away. Are you sure that you waited long enough for this to be freed?"
It's not going away - in one example, I sequentially opened and closed the same module 15 times, which took me more than 15mins to complete, and at the end all of the 15 eSpace references were still being held and 1.25GB of memory  was being used.

"In case that there's really a problem, clear all the handlers inside the pipe server instead."
I don't quite understand how I can clear the handlers inside the PipeServer - I need your insights on this one.

CefGlue.Common/CommonBrowserAdapter.cs
brunocunhasilva: After a Zoom chat we decided to remove these event unsubscriptions

CefGlue.Common/CommonBrowserAdapter.cs
brunocunhasilva: The MessageReceived handler is now being released in the PipeServer class and not in this one.

CefGlue.Common/CommonBrowserAdapter.cs
```diff
@@ -459,6 +461,11 @@ protected virtual bool OnBrowserClose(CefBrowser browser)
 
         protected void Cleanup(CefBrowser browser)
         {
+            if (_crashServerPipe != null)
```
brunocunhasilva: For reference, there's a bug in .NET Core 5 (fixed in .NET Core 6) that happens in Unix systems and that doesn't release the registered NamedPipeServerStreams - "WaitForConnectionAsync does not respond to cancellation"
https://github.com/dotnet/runtime/issues/40289

CefGlue.Common.Shared/RendererProcessCommunication/PipeServer.cs
```diff
@@ -38,12 +38,18 @@ public PipeServer(string pipeName)
...
+                var cancellationTokenSource = _cancellationTokenSource;
+                _cancellationTokenSource = null;
+                cancellationTokenSource.Dispose();
             });
         }
 
         public void Dispose()
         {
-            _cancellationTokenSource.Cancel();
+            _cancellationTokenSource?.Cancel();
```
brunocunhasilva, line 50: For reference, there's a bug in .NET Core 5 (fixed in .NET Core 6) that happens in Unix systems and that doesn't release the registered NamedPipeServerStreams - "WaitForConnectionAsync does not respond to cancellation"
https://github.com/dotnet/runtime/issues/40289

joaompneves (approved):

PR #51, opened 2022-06-29
Upgrade to avalonia 10.0.15

joaompneves:


alvesmiguel1 (dismissed):
Thanks for doing this change!

PR #54, opened 2022-09-21
Preserve references in serialization to allow deeper structures to be serialized

brunocunhasilva:
- [x] Preserve references
- [x] Fix failing tests
- [x] Create new tests
- [x] Create tests for the Validate revive() logic


CefGlue.BrowserProcess/ObjectBinding/JavascriptHelper.cs
```diff
@@ -24,14 +24,40 @@ internal static partial class JavascriptHelper
...
-            "                    case \"" + DataMarkers.DateTimeMarker + "\":" +
-            "                        return new Date(value.substring(" + DataMarkers.MarkerLength + "));" +
-            "                    case \"" + DataMarkers.BinaryMarker + "\":" +
-            "                        return convertBase64ToBinary(value.substring(" + DataMarkers.MarkerLength + "));" +
+            "            const refs = new Map();" +
+            "            const pendingRefs = new Map();" +
+            "            if (value) {" +
+            "                const id = value.$id;" +
+            "                if (id !== undefined) {" +
+            "                    delete value.$id;" +
```
joaompneves, line 32:
```suggestion
            "                if (value.$id !== undefined) {" +
            "                    const id = value.$id;" +
            "                    delete value.$id;" +
```

CefGlue.BrowserProcess/ObjectBinding/JavascriptHelper.cs
```diff
@@ -24,14 +24,40 @@ internal static partial class JavascriptHelper
...
+            "                    if (ref) {" +
+            "                        return ref;" +
+            "                    }" +
+            "                    const pendingRef = pendingRefs.get(refId);" +
+            "                    if (!pendingRef) {" +
+            "                        value = {};" +
+            "                        pendingRefs.set(refId, value);" +
+            "                    } else {" +
+            "                        value = pendingRef;" +
+            "                    }" +
```
joaompneves, line 51:
```suggestion
            "                    if (pendingRef) {" +
            "                        value = pendingRef;" +
            "                    } else {" +
            "                        value = {};" +
            "                        pendingRefs.set(refId, value);" +
            "                    }" +
```

CefGlue.BrowserProcess/ObjectBinding/JavascriptHelper.cs
```diff
@@ -24,14 +24,40 @@ internal static partial class JavascriptHelper
...
-            "            if (isString(value)) {" +
-            "                switch (value.substring(0, " + DataMarkers.MarkerLength + ")) {" +
-            "                    case \"" + DataMarkers.StringMarker + "\":" +
-            "                        return value.substring(" + DataMarkers.MarkerLength + ");" +
-            "                    case \"" + DataMarkers.DateTimeMarker + "\":" +
-            "                        return new Date(value.substring(" + DataMarkers.MarkerLength + "));" +
-            "                    case \"" + DataMarkers.BinaryMarker + "\":" +
-            "                        return convertBase64ToBinary(value.substring(" + DataMarkers.MarkerLength + "));" +
+            "            const refs = new Map();" +
+            "            const pendingRefs = new Map();" +
```
joaompneves, line 28:
```suggestion
            "        function revive(name, value, refs, pendingRefs) {" +
```

joaompneves:
```
            "        function parseResult(result) {" +
            "            const refs = new Map();" +
            "            const pendingRefs = new Map();" +
            "            return isString(result) ? JSON.parse(result, (name, value) => revive(name, value, refs, pendingRefs)) : result;"+
            "        }

File: CefGlue.Common.Shared/Serialization/CefValueSerialization.cs [hunk 4]
```diff
@@ -2,34 +2,37 @@
...
+using System.Text.Json.Serialization;
 
 namespace Xilium.CefGlue.Common.Shared.Serialization
 {
     internal static class CefValueSerialization
     {
-        private const int SerializerMaxDepth = byte.MaxValue;
-        
+        private const int SerializerMaxDepth = int.MaxValue;
+        private const int DeserializerMaxDepth = byte.MaxValue;
```
joaompneves, line 12: theres no connection with byte... so lets use a const
```suggestion
        private const int DeserializerMaxDepth = 255;
```

CefGlue.Tests/Serialization/CefValueSerializationTests.cs
```diff
@@ -214,6 +214,38 @@ public void HandlesDeepStructuresWith250Levels()
             AssertSerialization(list, CefValueType.String);
         }
 
+        [Test]
+        public void HandlesDeepStructuresWithReferenceObjects()
```
joaompneves, line 218: make a test with a single purpose ... reference... not need to test deep structures here.. already done on other tests

joaompneves (changes requested):
Also needs tests for the JS part of the changes.. which could be inspired on: the NativeObjectInteropTests

CefGlue.Common.Shared/Serialization/CefValueSerialization.cs
```diff
@@ -1,35 +1,55 @@
...
         private static readonly JsonSerializerOptions _jsonDeserializerOptions = new JsonSerializerOptions()
         {
             Converters =
             {
                 new ObjectJsonConverter()
             },
-            MaxDepth = SerializerMaxDepth
+            IncludeFields = true,
+            MaxDepth = DeserializerMaxDepth,
+            ReferenceHandler = _jsonDeserializerReferenceHandler
```
joaompneves, line 38: deserialization does not support references because cef because of this:
https://github.com/OutSystems/CefGlue/blob/4472/CefGlue.BrowserProcess/Serialization/V8ValueSerialization.cs#L24

CefGlue.Common.Shared/Serialization/CefValueSerialization.cs
```diff
@@ -1,35 +1,55 @@
...
+        private static readonly JsonReferenceHandler _jsonDeserializerReferenceHandler = new JsonReferenceHandler();
+
         private static readonly JsonSerializerOptions _jsonDeserializerOptions = new JsonSerializerOptions()
         {
             Converters =
             {
                 new ObjectJsonConverter()
             },
-            MaxDepth = SerializerMaxDepth
+            IncludeFields = true,
```
joaompneves, line 36: not needed

CefGlue.Common.Shared/Serialization/CefValueSerialization.cs
```diff
@@ -1,35 +1,55 @@
...
+
+        private static JsonSerializerOptions JsonDeserializerOptions
+        {
+            get
+            {
+                _jsonDeserializerReferenceHandler.Reset();
+                _jsonDeserializerOptions.Converters.OfType<ObjectJsonConverter>().First().ResetState();
+                return _jsonDeserializerOptions;
+            }
+        }
```
joaompneves, line 51: this is very dangerous since its not thread safe and with current system.text.json capabilities its impossible to solve

CefGlue.Common.Shared/Serialization/CefValueSerialization.cs
```diff
@@ -1,35 +1,55 @@
...
                 new ObjectJsonConverter()
             },
-            MaxDepth = SerializerMaxDepth
+            IncludeFields = true,
+            MaxDepth = DeserializerMaxDepth,
+            ReferenceHandler = _jsonDeserializerReferenceHandler
         };
-        
+
+        private static JsonSerializerOptions JsonSerializerOptions => _jsonSerializerOptions;
```
joaompneves, line 41: remove

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
```diff
@@ -7,6 +7,13 @@ namespace Xilium.CefGlue.Common.Shared.Serialization
 {
     internal class ObjectJsonConverter : JsonConverter<object>
```
joaompneves, line 8: lets not go this path... this has serious thread safety issues... concurrent deserializations will share the same references dictionary... 

to properly support cyclic references on deserialization, we would have to ditch the ObjectJsonConverter (because https://github.com/dotnet/runtime/issues/51715) and do the deserialization by ourselves. that shouldn't be hard to do, but I would put in another separated PR, since it involves some of changes.

joaompneves (changes requested):
Leave just the cyclic support on the serialization... in this PR

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
brunocunhasilva: What if we have the Deserializer options as threadstatic? (do you foresee problems in this approach)
image

joaompneves: there's no guarantees that the threads are not reused by the thread pool... while performing async work
also creating new json serializer options leads not only to performance issues (because internal caches have to be built from scratch for each instance, and also leads to memory leaks)

brunocunhasilva: I believe that the threads may be reused without any problem, since those threads will only pick up after the initial job has been completed, hence not sharing the references collection, since I can reset it for subsequent uses:
image

image


As for the potential memory leaks, if I reset the _jsonDeserializerReferenceHandler at the end of the Deserialization process, I don't think we have that problem either:
image

We're left with the potential performance issues - can you detail which internal caches are you referring to?

brunocunhasilva: Discarded this approach after aligning with Neves.
A new approach will be followed in a separate PR

CefGlue.Common.Shared/Serialization/CefValueSerialization.cs
```diff
@@ -1,35 +1,39 @@
 using System;
 using System.Collections.Generic;
 using System.Dynamic;
+using System.Linq;
```
joaompneves, line 4: unused

CefGlue.Tests/Javascript/NativeObjectInteropTests.cs
```diff
@@ -218,6 +232,25 @@ public async Task MethodWithObjectParamIsPassed()
...
+        public async Task MethodWithCyclicObjectParamIsPassed()
+        {
+            var taskCompletionSource = new TaskCompletionSource<object[]>();
+            nativeObject.MethodWithCyclicObjectParamCalled += (args) => taskCompletionSource.SetResult(args);
+
+            Execute($"{ObjName}.methodWithCyclicObjectParam({{'$id':'1','Name': 'parent1','Parent':null,'Child':{{'$id':'2','Name': 'child1','Parent':{{'$ref':'1'}},'Child':null}}}})");
+
+            var result = await taskCompletionSource.Task;
+            Assert.AreEqual(1, result.Length);
+            Assert.AreEqual(typeof(CyclicObject), result[0].GetType());
```
joaompneves, line 245: Assert.IsInstanceOfType

CefGlue.Tests/Javascript/NativeObjectInteropTests.cs
```diff
@@ -218,6 +232,25 @@ public async Task MethodWithObjectParamIsPassed()
...
+
+            var arg = (CyclicObject)result[0];
+            Assert.AreEqual("parent1", arg.Name);
+            Assert.NotNull(arg.Child);
+            Assert.AreEqual("child1", arg.Child.Name);
+            Assert.NotNull(arg.Child.Parent);
```
joaompneves, line 251: missing the reference equality check

CefGlue.Tests/Serialization/CefValueSerializationTests.cs
```diff
@@ -1,26 +1,66 @@
...
+            IncludeFields = true,
+            MaxDepth = 255,
+            ReferenceHandler = ReferenceHandler.Preserve
+        };
+
         private static object SerializeAndDeserialize(object value, out CefValueType valueType)
+        {
+            return SerializeAndDeserialize<object>(value, out valueType);
+        }
+        private static object SerializeAndDeserialize<T>(object value, out CefValueType valueType, DeserializationType deserializationType = DeserializationType.CefValue)
```
joaompneves, line 39:
```suggestion

        private static object SerializeAndDeserialize<T>(object value, out CefValueType valueType, DeserializationType deserializationType = DeserializationType.CefValue)
```

CefGlue.Tests/Serialization/CefValueSerializationTests.cs
```diff
@@ -134,23 +174,101 @@ public void HandlesDateTimes()
...
+            var list = new List<object>();
+            list.Add(list);
+
+            object obtainedValue = null;
+            Assert.DoesNotThrow(() => obtainedValue = SerializeAndDeserialize<List<object>>(list, out var _, DeserializationType.Standard));
+            Assert.IsInstanceOf<List<object>>(obtainedValue);
+            Assert.AreSame(obtainedValue,((List<object>)obtainedValue).First());
+        }
+
+        private class Person
```
joaompneves, line 200: move to the top

CefGlue.Tests/Serialization/CefValueSerializationTests.cs
```diff
@@ -134,23 +174,101 @@ public void HandlesDateTimes()
...
+            //     ((Dictionary<string, Object>)values[2]).Values.ToArray()[1], "Child.Parent instance should point to the obtained dictionary instance.");
         }
 
         [Test]
         public void HandlesLists()
         {
             var list = new List<string>() { "1", "2" };
-            AssertSerialization(list, CefValueType.String);
-            AssertSerialization(new List<string>(), CefValueType.String);
+            AssertSerialization<List<string>>(list, CefValueType.String, DeserializationType.Standard);
```
joaompneves, line 268: needs generics to be explicitly passed?

CefGlue.Tests/Serialization/CefValueSerializationTests.cs
```diff
@@ -176,42 +294,79 @@ public void HandlesListsOfObjects()
...
+
             var cefValue = new CefTestValue();
-            Assert.Throws<InvalidOperationException>(() => Serialize(list, cefValue));
+            Assert.DoesNotThrow(() => Serialize(list, cefValue));
         }
-        
+
         [Test]
-        public void HandlesDeepStructuresWith250Levels()
+        public void HandlesDeserializationOfDeepListsWith248LevelsWithoutReferences()
```
joaompneves, line 327: whats this case for? how is this different from the one above?

CefGlue.Tests/Serialization/CefValueSerializationTests.cs
```diff
@@ -176,42 +294,79 @@ public void HandlesListsOfObjects()
...
+            object obtainedValue = null;
+
+            Assert.DoesNotThrow(() => obtainedValue = DeserializeCefValue(cefValue));
+            var valueType = cefValue.GetValueType();
+            Assert.IsTrue(valueType == CefValueType.String);
+            Assert.IsInstanceOf<object[]>(obtainedValue);
+        }
+
+        [Test]
+        public void HandlesReferenceObjects()
```
joaompneves, line 360: HandlesCyclicObjectReferences covers this

joaompneves (changes requested):
Lots of comments out code... move the commented out code to a new PR so that thus one remains clean

CefGlue.Tests/Serialization/CefValueSerializationTests.cs
```diff
@@ -134,23 +174,101 @@ public void HandlesDateTimes()
...
+            var parent = new Person("p");
+            var child = new Person("c");
+            child.Parent = parent;
+            parent.Child = child;
+
+            AssertSerializeAndDeserializeOfCyclicObjectReferences(parent);
+        }
+
+        [Test]
+        public void Handles100ParallelTasksOfCyclicObjectReferences()
```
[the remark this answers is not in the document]
brunocunhasilva: shouldn't we keep it to avoid anyone, in the future, introducing state that may break multi-threaded serialization?

CefGlue.Tests/Javascript/NativeObjectInteropTests.cs
brunocunhasilva: Considering that the JS still doesn't serialize the objects with $id, $ref properties, the equality check can't currently be done.

CefGlue.Tests/Serialization/CefValueSerializationTests.cs
brunocunhasilva: Yes - otherwise the resulting type won't be the same - it'll be an array

CefGlue.Tests/Serialization/CefValueSerializationTests.cs
brunocunhasilva: This one specifically tests the DeserializeCefValue, while the above one tests the Serialize only.

CefGlue.Tests/Serialization/CefValueSerializationTests.cs
brunocunhasilva: Yes, otherwise the obtained type will be an array and won't match the original type.

CefGlue.Tests/Serialization/CefValueSerializationTests.cs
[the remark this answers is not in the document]
joaompneves: nah... there might be ways to break taht that this test wont cover.. and also multi-threading tests are always unpredictable

CefGlue.Tests/Serialization/CefValueSerializationTests.cs
```diff
@@ -1,26 +1,79 @@
...
+            },
+            IncludeFields = true,
+            MaxDepth = 255,
+            ReferenceHandler = ReferenceHandler.Preserve
+        };
+
         private static object SerializeAndDeserialize(object value, out CefValueType valueType)
+        {
+            return SerializeAndDeserialize<object>(value, out valueType);
+        }
```
joaompneves, line 50: useless method

CefGlue.Tests/Serialization/CefValueSerializationTests.cs
joaompneves: why 248?

CefGlue.Tests/Serialization/StringJsonConverter.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Xilium.CefGlue.Common.Shared.Serialization;
+
+namespace CefGlue.Tests.Serialization
+{
+    internal class StringJsonConverter : JsonConverter<string>
```
joaompneves, line 9: this is temporary?

CefGlue.Tests/Serialization/CefValueSerializationTests.cs
brunocunhasilva: It was a left over from the changes done to ObjectJsonSerializer that were able to handle 248 levels. I'll change to 250, to match the Serializer test.

CefGlue.Tests/Serialization/CefValueSerializationTests.cs
brunocunhasilva: I'll leave it, because it'll make it easier to refactor the tests in the [#55](https://github.com/OutSystems/CefGlue/pull/55) PR, since the method with the Generic won't be needed anymore.

CefGlue.Tests/Serialization/StringJsonConverter.cs
brunocunhasilva: Yes - it won't be needed when the [#55 PR](https://github.com/OutSystems/CefGlue/pull/55) is complete.

CefGlue.Tests/Serialization/CefValueSerializationTests.cs
```diff
@@ -211,7 +299,34 @@ public void HandlesDeepStructuresWith250Levels()
...
+        public void HandlesListWithObjectReferences()
+        {
+            var list = new List<object>();
+            var childList = new List<object>();
+            list.Add(childList);
+            list.Add(childList);
+
+            var cefValue = new CefTestValue();
+            Assert.DoesNotThrow(() => Serialize(list, cefValue));
+            Assert.IsTrue(cefValue.GetString().TrimEnd('}', ']').EndsWith("\"$ref\":\"2\""), "The last element in the list should be a reference to the first child");
```
joaompneves, line 329: just a note... when deserialization comes in these tests should stop comparing strings

CefGlue.Tests/Serialization/CefValueSerializationTests.cs
joaompneves: its not used

CefGlue.Tests/Serialization/CefValueSerializationTests.cs
```diff
@@ -134,23 +182,55 @@ public void HandlesDateTimes()
...
+        public void HandlesCyclicDictionaryReferences()
         {
             var dict = new Dictionary<string, object>();
             dict.Add("first", dict);
-            
-            var cefValue = new CefTestValue();
-            Assert.Throws<InvalidOperationException>(() => Serialize(dict, cefValue));
+
+            object obtainedValue = null;
+            Assert.DoesNotThrow(() => obtainedValue = SerializeAndDeserialize<Dictionary<string, object>>(dict, out var _, DeserializationType.Standard));
```
joaompneves, line 191:
```suggestion
            Assert.DoesNotThrow(() => obtainedValue = SerializeAndDeserialize(dict, out var _, DeserializationType.Standard));
```

CefGlue.Tests/Serialization/CefValueSerializationTests.cs
```diff
@@ -134,23 +182,55 @@ public void HandlesDateTimes()
...
+        }
+
+        [Test]
+        public void HandlesCyclicListReferences()
+        {
+            var list = new List<object>();
+            list.Add(list);
+
+            object obtainedValue = null;
+            Assert.DoesNotThrow(() => obtainedValue = SerializeAndDeserialize<List<object>>(list, out var _, DeserializationType.Standard));
```
joaompneves, line 203:
```suggestion
            Assert.DoesNotThrow(() => obtainedValue = SerializeAndDeserialize(list, out var _, DeserializationType.Standard));
```

CefGlue.Tests/Serialization/CefValueSerializationTests.cs
```diff
@@ -134,23 +182,55 @@ public void HandlesDateTimes()
...
+        [Test]
+        public void HandlesCyclicObjectReferences()
+        {
+            var parent = new Person("p");
+            var child = new Person("c");
+            child.Parent = parent;
+            parent.Child = child;
+
+            object obtainedValue = null;
+            Assert.DoesNotThrow(() => obtainedValue = SerializeAndDeserialize<Person>(parent, out var _, DeserializationType.Standard));
```
joaompneves, line 217:
```suggestion
            Assert.DoesNotThrow(() => obtainedValue = SerializeAndDeserialize(parent, out var _, DeserializationType.Standard));
```

PR #55, opened 2022-09-28
Preserve references in deserialization

brunocunhasilva:
- [x] Replace ObjectJsonConverter
- [x] Add logic to preserve references
- [x] Support references in JS serialization
- [x] Refactor tests


CefGlue.Tests/Serialization/CefValueSerializationTests.cs
```diff
@@ -338,13 +333,7 @@ public void HandlesSimpleDictionaries()
                 { "seventh", 7.0 }
             };
```
joaompneves, line 335: change also https://github.com/OutSystems/CefGlue/pull/55/files#diff-519d5dbffd5b99ec72bc636a0577d3d58bfe8139ef3de35539c2276fd4e2c53bR319

CefGlue.Tests/Serialization/CefValueSerializationTests.cs
```diff
@@ -262,6 +243,20 @@ public void HandlesArrays()
             AssertSerialization(new string[0], CefValueType.String);
```
joaompneves, line 243: also check the return type of the deserialized object... should be an array

CefGlue.Tests/Serialization/CefValueSerializationTests.cs
```diff
@@ -209,21 +184,27 @@ public void HandlesCyclicObjectReferences()
...
         }
 
         [Test]
         public void HandlesLists()
         {
             var list = new List<string>() { "1", "2" };
-            AssertSerialization(list, CefValueType.String, DeserializationType.Standard);
-            AssertSerialization(new List<string>(), CefValueType.String, DeserializationType.Standard);
+            AssertSerialization(list, CefValueType.String);
+            AssertSerialization(new List<string>(), CefValueType.String);
```
joaompneves, line 207: also check the return type of the deserialized object... should be an array

CefGlue.Tests/Serialization/CefValueSerializationTests.cs
```diff
@@ -209,21 +184,27 @@ public void HandlesCyclicObjectReferences()
...
+            Assert.AreEqual(3, ((Dictionary<string, object>)obtainedValue).Count);
+            var keys = ((Dictionary<string, object>)obtainedValue).Keys.ToArray();
+            Assert.AreEqual("Parent", keys[1]);
+            Assert.AreEqual("Child", keys[2]);
+            var values = ((Dictionary<string, object>)obtainedValue).Values.ToArray();
+            Assert.AreEqual(parent.Name,
+                values[0]);
+            Assert.AreEqual(parent.Child.Name,
+                ((Dictionary<string, Object>)values[2]).Values.First());
+            Assert.AreSame(obtainedValue,
```
joaompneves, line 198: ToArray ... that wont work...it will never be the same reference

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
```diff
@@ -1,62 +1,172 @@
...
+            }
+            else if (obj is object[] array)
+            {
+                var arrayIndex = arraysIndexesStack.Pop();
+                array[arrayIndex] = value;
+                arraysIndexesStack.Push(++arrayIndex);
             }
         }
 
+        private int GetArraySize(Utf8JsonReader reader)
```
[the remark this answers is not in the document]
brunocunhasilva: Code was refactored to be simpler

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
```diff
@@ -1,60 +1,207 @@
...
+            referencesMap.Add(objId, value);
+            return true;
+        }
+
+        private int PeekAndCalculateArraySize(Utf8JsonReader reader)
+        {
+            var arraySize = 0;
+            var startDepth = reader.CurrentDepth;
+            reader.ReadToken(JsonTokenType.StartArray);
+            while (reader.CurrentDepth > startDepth || reader.TokenType != JsonTokenType.EndArray)
```
joaompneves, line 192: I think it reads better because the stop condition is when the current depth == start and we found the endarray token
```suggestion
            while (!(reader.CurrentDepth == startDepth && reader.TokenType == JsonTokenType.EndArray))
```

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
```diff
@@ -1,60 +1,207 @@
...
+            {
+                if (reader.CurrentDepth == startDepth + 1
+                    && reader.TokenType != JsonTokenType.EndArray
+                    && reader.TokenType != JsonTokenType.EndObject
+                    && reader.TokenType != JsonTokenType.PropertyName
```
joaompneves, line 197: not needed ... that would be malformed json

CefGlue.Common.Shared/Serialization/JsonReaderExtensions.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System;
+using System.Text.Json;
+
+namespace Xilium.CefGlue.Common.Shared.Serialization;
+
+public static class JsonReaderExtensions
```
joaompneves, line 6: are you using all these extension methods?

CefGlue.Common.Shared/Serialization/JsonReaderExtensions.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System;
+using System.Text.Json;
+
+namespace Xilium.CefGlue.Common.Shared.Serialization;
```
joaompneves, line 4: file formmating not compliant with cefglue

CefGlue.Common/Helpers/JavascriptToNativeTypeConverter.cs
```diff
@@ -31,6 +31,18 @@ public static T ConvertToNative<T>(object obj)
...
+        }
+
+        /// <summary>
+        /// Converts an object from javascript into a .Net type
+        /// </summary>
+        /// <param name="obj"></param>
+        /// <param name="expectedType">Expected object type</param>
+        /// <param name="convertedObjectsMap">Dictionary of the objects (key) that were converted to NativeObjects (values)</param>
+        /// <returns></returns>
+        public static object ConvertToNative(object obj, Type expectedType, Dictionary<object, object> convertedObjectsMap)
```
joaompneves, line 45:
```suggestion
        public static object ConvertToNative(object obj, Type expectedType, IDictionary<object, object> convertedObjectsMap)
```

joaompneves: needs to be public?

CefGlue.Common/Helpers/JavascriptToNativeTypeConverter.cs
```diff
@@ -130,9 +142,15 @@ private static object ConvertToNativeList(object obj, Type expectedType)
             return nativeList;
         }
 
-        private static object ConvertToNativeObject(object obj, Type expectedType)
+        private static object ConvertToNativeObject(object obj, Type expectedType, Dictionary<object, object> convertedObjectsMap)
```
joaompneves, line 145:
```suggestion
        private static object ConvertToNativeObject(object obj, Type expectedType, IDictionary<object, object> convertedObjectsMap)
```

CefGlue.Common/Helpers/JavascriptToNativeTypeConverter.cs
```diff
@@ -74,7 +86,7 @@ public static object ConvertToNative(object obj, Type expectedType)
                 return ConvertToNativeList(obj, expectedType);
```
joaompneves, line 86: what about lists?

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
```diff
@@ -1,60 +1,207 @@
...
+            }
+            else
+            {
+                value = new Dictionary<string, object>();
+            }
+            referencesMap.Add(objId, value);
+            return true;
+        }
+
+        private int PeekAndCalculateArraySize(Utf8JsonReader reader)
```
joaompneves, line 187: this should go to the json extensions

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
```diff
@@ -1,60 +1,207 @@
...
+            var isId = propName == JsonAttributeIdPropName;
+            var isRef = !isId && propName == JsonAttributeRefPropName;
+            if (!isId && !isRef)
+            {
+                value = null;
+                return false;
+            }
+
+            reader.Read(); // StartObject
+            reader.Read(); // PropertyName
```
joaompneves, line 159: replace with read tokens

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
```diff
@@ -1,60 +1,207 @@
...
 
-                case JsonTokenType.True:
-                    return true;
+            public readonly bool HasReference;
+            public readonly object ObjectHolder;
 
-                case JsonTokenType.False:
-                    return false;
+            public string PropertyName;
+            public int? ArrayIndex;
```
joaompneves, line 29: should be props

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
```diff
@@ -1,60 +1,207 @@
...
     {
-        public override object Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        internal const string JsonAttributeIdPropName = "$id";
+        internal const string JsonAttributeRefPropName = "$ref";
+        private const string JsonAttributeValuesPropName = "$values";
+
+        private class ReadState
         {
-            switch (reader.TokenType)
+            public ReadState(string propertyName, int? arrayIndex, bool hasReference, object objectHolder)
```
joaompneves, line 17: doesnt make sense to have all these params simultaneously in a single ctor

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
```diff
@@ -1,60 +1,207 @@
...
-                    return null;
+                PropertyName = propertyName;
+                ArrayIndex = arrayIndex;
+                HasReference = hasReference;
+                ObjectHolder = objectHolder;
+            }
 
-                case JsonTokenType.True:
-                    return true;
+            public readonly bool HasReference;
```
joaompneves, line 25: whats this for?

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
```diff
@@ -1,60 +1,207 @@
...
+                        break;
 
-                case JsonTokenType.StartObject:
-                    return JsonDocument.ParseValue(ref reader).Deserialize<Dictionary<string, object>>(options);
+                    case JsonTokenType.EndArray:
+                        if (!currentState.HasReference)
+                        {
+                            stateStack.Pop();
+                        }
+                        continue;
```
joaompneves, line 101: these cases should no longer be applicable

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
```diff
@@ -1,60 +1,207 @@
...
+                ((Dictionary<string, object>)obj)[state.PropertyName] = value;
+            }
+            else if (state.ArrayIndex != null)
+            {
+                ((object[])obj)[state.ArrayIndex.Value] = value;
+                state.ArrayIndex++;
+            }
+        }
+
+        private bool HandleReferences(ref Utf8JsonReader reader, Dictionary<string, object> referencesMap, out object value, out bool isNewArray)
```
joaompneves, line 143: this method logic is confusing..

joaompneves (changes requested):

CefGlue.Tests/Serialization/CefValueSerializationTests.cs
brunocunhasilva: It's not comparing the ToArray instances, but the specific array elements inside of the array. And they are the same instance.

CefGlue.Tests/Serialization/CefValueSerializationTests.cs
brunocunhasilva: Didn't understand what you're suggesting here.

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
brunocunhasilva: They still are, because the RendererProcessCommunication still deserializes plain arrays, such as it happens here:
https://github.com/OutSystems/CefGlue/blob/adcb6d5557d9134e4041032f50fe0341f1a8ef72/CefGlue.Common.Shared/RendererProcessCommunication/Messages.cs#L109

Moreover, not supporting this would make us over-engineer the SerializationTests, so that we could validate the object[] serializations without using the deserializer.

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
brunocunhasilva: Let's have a zoom chat and discuss how we can improve this

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
```diff
@@ -1,60 +1,331 @@
...
+                switch (reader.TokenType)
+                {
+                    case JsonTokenType.Null:
+                        value = null;
+                        break;
 
-                case JsonTokenType.True:
-                    return true;
+                    case JsonTokenType.Number:
+                        value = reader.TryGetInt64(out var number)
```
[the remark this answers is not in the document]
brunocunhasilva: Using the GetNumber instead of Read, as we would lose readability in the cycle.

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
```diff
@@ -1,60 +1,331 @@
...
+                            ? number
+                            : (object)reader.GetDouble();
+                        break;
+
+                    case JsonTokenType.String:
+                        var stringValue = reader.GetString();
```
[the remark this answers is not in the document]
brunocunhasilva: Same as for GetNumber

CefGlue.Tests/Serialization/CefValueSerializationTests.cs
brunocunhasilva: being done inside of the AssertSerialization method, by converting the passed list to an array and comparing it with the deserialize result:
image

CefGlue.Common.Shared/Serialization/CefReferencesResolver.cs
```diff
@@ -0,0 +1,83 @@
...
+            }
+
+            return value;
+        }
+
+        public bool TryGetReferenceId(T value, out string referenceId)
+        {
+            if (_equalityComparer != null)
+            {
+                var key = ObjectToReferenceIdMap.Keys.FirstOrDefault(k => _equalityComparer.Equals(k, value));
```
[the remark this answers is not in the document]
brunocunhasilva: This is how I was able to solve the problem of searching for the CefV8Values in the map, considering that equal V8CefValues may have different HashCodes. Can you suggest a better approach?

CefGlue.Common.Shared/Serialization/CefValueSerialization.cs
```diff
@@ -138,6 +138,11 @@ private static void SerializeAsJson(object value, CefValueWrapper cefValue)
         }
 
         public static object DeserializeCefValue(CefValueWrapper cefValue)
+        {
+            return DeserializeCefValue(cefValue, new CefReferencesResolver<object>());
+        }
+
+        public static object DeserializeCefValue(CefValueWrapper cefValue, IReferencesResolver<object> referencesResolver)
```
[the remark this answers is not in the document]
brunocunhasilva: This was already `public`, but yes, it can be reduced to `internal`

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
```diff
@@ -1,60 +1,207 @@
...
+                                    value = JsonSerializer.Deserialize<DateTime>("\"" + stringValue.Substring(DataMarkers.MarkerLength) + "\"");
+                                    break;
 
-                            case DataMarkers.BinaryMarker:
-                                return Convert.FromBase64String(value.Substring(DataMarkers.MarkerLength));
+                                case DataMarkers.BinaryMarker:
+                                    value = Convert.FromBase64String(stringValue.Substring(DataMarkers.MarkerLength));
+                                    break;
+                            }
                         }
```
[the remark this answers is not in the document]
brunocunhasilva: Created an extension for this - `GetObjectFromString`

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
brunocunhasilva: I actually find it harder to understand as it's a negation of a multiple steps condition. The current resolves in a simpler way in my mid:
Continue if condition 1 is true or condition 2 is true

CefGlue.Common/Helpers/JavascriptToNativeTypeConverter.cs
brunocunhasilva: Didn't understand what you mean

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
brunocunhasilva: I can only remove the propertyName from the constructor, considering it's not being used

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
brunocunhasilva: This is needed to distinguish from arrays with reference id - `$id=N,$values=[...]` from plain arrays [].

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
```diff
@@ -1,60 +1,171 @@
...
+                        stateStack.Push(new ReadState(0, false, value));
+                        break;
 
-                            case DataMarkers.DateTimeMarker:
-                                return JsonSerializer.Deserialize<DateTime>("\"" + value.Substring(DataMarkers.MarkerLength) + "\"");
+                    case JsonTokenType.EndArray:
+                        if (!currentState.HasReference)
+                        {
+                            stateStack.Pop();
+                        }
```
joaompneves, line 79:
```suggestion
                        stateStack.Pop();
```

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
```diff
@@ -1,60 +1,171 @@
...
+
+            public bool HasReference { get; }
+
+            public object ObjectHolder { get; }
+            
+            public string PropertyName { get; set; }
+            
+            public int? ArrayIndex { get; set; }
+        }
+
```
joaompneves, line 28: private static readonly ReadState ListWrapperMarker = new ReadState();

CefGlue.Common.Shared/Serialization/CefValueSerialization.cs
```diff
@@ -168,7 +199,7 @@ public static object DeserializeCefValue(CefValueWrapper cefValue)
                 case CefValueType.List:
                     using (var cefList = cefValue.GetList())
                     {
-                        return DeserializeCefList<object>(cefList);
+                        return DeserializeCefList<object>(cefList, referencesResolver);
```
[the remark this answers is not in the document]
brunocunhasilva: We need to pass the resolver here, because the List may appear nested inside a other dictionaries - if we don't pass the referencesResolver, the list owner dictionary references can't be resoved

CefGlue.Common.Shared/Serialization/ReferencesResolver.cs
```diff
@@ -0,0 +1,82 @@
...
+using System.Text;
+using System.Text.Json;
+using System.Threading.Tasks;
+
+namespace Xilium.CefGlue.Common.Shared.Serialization
+{
+    internal class ReferencesResolver<T> : IReferencesResolver<T>
+    {
+        private Lazy<Dictionary<string, T>> _lazyReferenceIdToObjectMap;
+        private Lazy<Dictionary<T, string>> _lazyObjectToReferenceIdMap;
```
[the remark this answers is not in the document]
brunocunhasilva: Wasn't this what you intended by [this](https://github.com/OutSystems/CefGlue/pull/55#discussion_r994811965) comment?

CefGlue.Common.Shared/Serialization/CefValueSerialization.cs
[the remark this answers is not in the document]
joaompneves: yeah... but like I said then the list also needs to be registered

CefGlue.Common.Shared/Serialization/ReferencesResolver.cs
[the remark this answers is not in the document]
joaompneves: yes, but only the _lazyObjectToReferenceIdMap would justify taht... either way i don't think that  the overhead justifies... I would create the dictionaries right away

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
```diff
@@ -1,60 +1,171 @@
...
 using System.Collections.Generic;
+using System.Linq;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 
 namespace Xilium.CefGlue.Common.Shared.Serialization
 {
     internal class ObjectJsonConverter : JsonConverter<object>
     {
+        private class ReadState
```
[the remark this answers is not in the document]
brunocunhasilva: We can't make full use of the record keyword, because CefGlue still targets .net472, and we get compiler errors when using certain patterns.
Check the answer in this [thread](https://developercommunity.visualstudio.com/t/error-cs0518-predefined-type-systemruntimecompiler/1244809)

CefGlue.Common.Shared/Serialization/CefValueSerialization.cs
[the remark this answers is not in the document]
brunocunhasilva: No - the list doesn't have a reference id - only the dictionaries (that correspond to objects) need to be registered.

CefGlue.Common.Shared/Serialization/CefValueSerialization.cs
```diff
@@ -202,14 +233,19 @@ private static object DeserializeAsJson(string value)
             }
         }
 
-        public static TListElementType[] DeserializeCefList<TListElementType>(ICefListValue cefList)
+        internal static TListElementType[] DeserializeCefList<TListElementType>(ICefListValue cefList, IReferencesResolver<object> referencesResolver = null)
```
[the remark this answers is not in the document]
brunocunhasilva: https://github.com/OutSystems/CefGlue/pull/55/files#r1030546677

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
```diff
@@ -1,60 +1,171 @@
...
+                        if (!currentState.HasReference)
+                        {
+                            stateStack.Pop();
+                        }
+                        continue;
 
-                            case DataMarkers.BinaryMarker:
-                                return Convert.FromBase64String(value.Substring(DataMarkers.MarkerLength));
+                    case JsonTokenType.StartObject:
+                        if (!HandleReferences(ref reader, referencesMap, out value, out var isNewArray))
```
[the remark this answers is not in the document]
brunocunhasilva: Had to do some changes to the proposed approach, to work as expected.

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
```diff
@@ -1,60 +1,171 @@
...
+            }
+            else if (state.ArrayIndex != null)
+            {
+                ((object[])obj)[state.ArrayIndex.Value] = value;
+                state.ArrayIndex++;
+            }
+        }
+
+        private bool HandleReferences(ref Utf8JsonReader reader, Dictionary<string, object> referencesMap, out object value, out bool isNewArray)
+        {
```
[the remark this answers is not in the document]
brunocunhasilva: Had to do some changes to the proposed approach, so it works as expected.

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
```diff
@@ -1,60 +1,170 @@
...
+                        if (tempReader.ReadPropertyName() == JsonAttributes.Values)
+                        {
+                            // it's a list
+                            reader.Read(); // skip the $id
+                            reader.ReadPropertyName(); // advance reader to the beginning of the list
+                            obj = new object[reader.PeekAndCalculateArraySize()];
+                            state.Push(ListWrapperMarker);
+                            newState = new ReadState(arrayIndex: 0, objectHolder: obj);
+                        }
+                        obj ??= new Dictionary<string, object>();
```
joaompneves, line 154:
```suggestion
                        }
                        else 
                        {
                            obj = new Dictionary<string, object>();
                        }
```

CefGlue.Common.Shared/Serialization/ReferencesResolver.cs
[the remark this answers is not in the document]
brunocunhasilva: I tend to disagree because there are many serializations&deserializations that don't require object references. Having the collections initialized for every serialization&deserialization will have some memory overhead that shouldn't be neglected.

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
```diff
@@ -1,60 +1,170 @@
...
+                        reader.ReadToken(JsonTokenType.StartObject);
+                        reader.ReadPropertyName();  // skip the $ref
+                        var refId = reader.GetString();
+                        obj = referencesMap[refId];
+                        break;
+
+                    case JsonAttributes.Id:
+                        reader.ReadToken(JsonTokenType.StartObject);
+                        reader.ReadPropertyName(); // skip the $id
+                        var id = tempReader.ReadString();
```
joaompneves, line 144: i guess its the reader here, at leats its not consistent with the branch above
```suggestion
                        var id = reader.ReadString();
```

CefGlue.BrowserProcess/Serialization/CefV8ValueEqualityComparer.cs
```diff
@@ -0,0 +1,27 @@
...
+        public static CefV8ValueEqualityComparer Instance => _instance;
+
+        public bool Equals(CefV8Value x, CefV8Value y)
+        {
+            return x.IsSame(y);
+        }
+
+        public int GetHashCode(CefV8Value obj)
+        {
+            return obj.GetHashCode();
```
joaompneves, line 24: this gives you the hashcode for the .net object... i think that this wont work as intended

CefGlue.Common.Shared/Serialization/ReferencesResolver.cs
```diff
@@ -0,0 +1,82 @@
...
+
+            throw new ArgumentException($"Reference not found - id={referenceId}");
+            
+        }
+
+        public bool TryGetReferenceId(T value, out string referenceId)
+        {
+            if (_equalityComparer != null)
+            {
+                var key = ObjectToReferenceIdMap.Keys.FirstOrDefault(k => _equalityComparer.Equals(k, value));
```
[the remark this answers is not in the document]
joaompneves: we need to investigate this further. this can be very ineficient. probably a live debugging session to see some other alternatives would help

CefGlue.Common.Shared/Serialization/ReferencesResolver.cs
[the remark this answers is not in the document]
joaompneves: ok, but in that case create a dummy implementation of IReferencesResolver that is basically a placebo

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
brunocunhasilva: It can't be the reader, because we're peeking for the $values property - if it's not, we can't consume that reader position, because it will be consumed in the while cycle continuation.

joaompneves: but something is not right, the branch above doesn't do the same... and they should be consistent because the scenario is very similar

CefGlue.BrowserProcess/Serialization/CefV8ValueEqualityComparer.cs
brunocunhasilva: It returns the value for the reference object and not a calculated value of the ValueObject, which is what would be ideal - but I simply don't know how to do it.

CefGlue.Common.Shared/Serialization/ReferencesResolver.cs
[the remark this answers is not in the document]
brunocunhasilva: Not possible, because you can't foresee if a regular implementation of IReferencesResolver will be needed or not. Only when you go through the deserialization process you are able to detect if the references resolver is needed or not. That's why we're instantiating it at this moment:
image

CefGlue.Common.Shared/Serialization/CefValueSerialization.cs
```diff
@@ -137,7 +138,7 @@ private static void SerializeAsJson(object value, CefValueWrapper cefValue)
             }
         }
 
-        public static object DeserializeCefValue(CefValueWrapper cefValue)
+        public static object DeserializeCefValue(CefValueWrapper cefValue, IReferencesResolver<object> referencesResolver = null)
```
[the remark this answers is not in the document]
brunocunhasilva: This way, we'll always instantiate a referencesResolver, even when it's not needed - like it's the case of arrays of value types.
Do you think the null check will have a relevant performance impact?

CefGlue.Common.Shared/Serialization/JsonConverters/ObjectJsonConverter.cs
brunocunhasilva: Not exactly the same as the branch above, because that branch uses the `reader.GetString()`, which doesn't consume the position. And if I were to use the `reader.GetString()`, I'd be forced to add another line just to consume that position in the tempReader so I could read the following PropertyName, to check if it's $values.

CefGlue.Common/Helpers/JavascriptToNativeTypeConverter.cs
brunocunhasilva: Got it. Added a couple of tests and corrected the code.

CefGlue.Common.Shared/CefGlue.Common.Shared.csproj
```diff
@@ -5,6 +5,7 @@
     <AssemblyName>Xilium.CefGlue.Common.Shared</AssemblyName>
     <RootNamespace>Xilium.CefGlue.Common.Shared</RootNamespace>
     <Configurations>Debug;Release;Debug_WindowlessRender</Configurations>
+    <LangVersion>latest</LangVersion>
```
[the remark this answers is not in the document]
brunocunhasilva: Wasn't able to. Doing that move from the csproj to the Directory.Build.props led to compilation errors.

CefGlue.Common.Shared/Serialization/CefValueSerialization.cs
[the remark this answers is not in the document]
brunocunhasilva: I ended up splitting the method into an `internal` (that doesn't expose the `referencesResolver`) and a `private` one that receives the `referencesResolver` argument

CefGlue.Common.Shared/Serialization/ReferencesResolver.cs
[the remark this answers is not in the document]
brunocunhasilva: Currently, the code already suffers from this inefficiency, in a more impactful way, considering that it always compares the V8Values in every serialization:
https://github.com/OutSystems/CefGlue/blob/e635050bfd356e8d80e3c2566fe0ff3018621858/CefGlue.BrowserProcess/Serialization/V8ValueSerialization.cs#L22

In this PR we're removing that check and reducing the comparisons that are made.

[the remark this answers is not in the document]
joaompneves: but it only compares a few objects (many as the nesting level)

CefGlue.Common.Shared/Serialization/CefReferencesResolver.cs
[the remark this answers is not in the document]
brunocunhasilva: This approach was removed. Closing the comment

CefGlue.Common.Shared/Serialization/ReferencesResolver.cs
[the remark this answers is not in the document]
brunocunhasilva: The new implemented solution no longer uses this approach. Closing this conversation

CefGlue.BrowserProcess/Serialization/CefV8ValueEqualityComparer.cs
brunocunhasilva: The new solution no longer uses this comparer. Closing this conversation

CefGlue.Common/InternalHandlers/CommonCefRenderHandler.cs
```diff
@@ -64,7 +64,7 @@ protected override void OnPaint(CefBrowser browser, CefPaintElementType type, Ce
         {
             if (_logger.IsDebugEnabled)
             {
-                _logger.Debug($"Type: {type} Buffer: {buffer.ToInt64()} Width: {width} Height: {height}");
+                _logger.Debug($"TypeInfo: {type} Buffer: {buffer.ToInt64()} Width: {width} Height: {height}");
```
joaompneves, line 67: intended?

CefGlue.Common.Shared/Serialization/State/TypeMethodInfo.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Xilium.CefGlue.Common.Shared.Serialization.State
+{
+    internal record TypeMethodInfo
```
joaompneves, line 9: records should be pure DTOs
```suggestion
    internal class TypeMethodInfo
```

CefGlue.Common.Shared/Serialization/Deserializer.cs
```diff
@@ -0,0 +1,275 @@
...
+using System.Text.Json;
+using Xilium.CefGlue.Common.Shared.Helpers;
+using Xilium.CefGlue.Common.Shared.Serialization.State;
+using static Xilium.CefGlue.Common.Shared.Serialization.State.ParametersDeserializerState;
+
+namespace Xilium.CefGlue.Common.Shared.Serialization
+{
+    internal class Deserializer
+    {
+        private const int DeserializerMaxDepth = byte.MaxValue * 2;
```
joaompneves, line 14: maxdepth can be the max value allowed... we no longer depend on the framework recursion limits

CefGlue.Common.Shared/Serialization/Deserializer.cs
```diff
@@ -0,0 +1,275 @@
...
+                    return default(TargetType);
+                }
+
+                return (TargetType)Deserialize(jsonString, typeof(TargetType), parametersTypes);
+            }
+            catch (JsonException e)
+            {
+                // wrap the json exception
+                throw new InvalidOperationException(e.Message);
+            }
```
joaompneves, line 53:
```suggestion
                            if (jsonString == null)
                {
                    return default(TargetType);
                }

            try
            {
                return (TargetType)Deserialize(jsonString, typeof(TargetType), parametersTypes);
            }
            catch (JsonException e)
            {
                // wrap the json exception
                throw new InvalidOperationException(e.Message);
            }
```

CefGlue.Common.Shared/Serialization/Deserializer.cs
```diff
@@ -0,0 +1,275 @@
...
+        }
+
+        private static object Deserialize(string jsonString, Type typeToConvert, ParametersTypes parametersTypes = null)
```
joaompneves, line 56:
```suggestion
        private static T InnerDeserialize<T>(string jsonString, ParametersTypes parametersTypes = null)
```

CefGlue.Common.Shared/Serialization/Deserializer.cs
```diff
@@ -0,0 +1,275 @@
...
+        {
+            byte[] bytes = Encoding.UTF8.GetBytes(jsonString);
+            var reader = new Utf8JsonReader(bytes, new JsonReaderOptions() { MaxDepth = DeserializerMaxDepth });
+
+            reader.Read();
+            return Deserialize(ref reader, typeToConvert, parametersTypes);
+        }
+
+        private static object Deserialize(ref Utf8JsonReader reader, Type typeToConvert, ParametersTypes parametersTypes = null)
```
joaompneves, line 65:
```suggestion
        private static T Deserialize<T>(ref Utf8JsonReader reader, ParametersTypes parametersTypes = null)
```

CefGlue.Common.Shared/Serialization/Deserializer.cs
```diff
@@ -0,0 +1,275 @@
...
+            catch (JsonException e)
+            {
+                // wrap the json exception
+                throw new InvalidOperationException(e.Message);
+            }
+        }
+
+        private static object Deserialize(string jsonString, Type typeToConvert, ParametersTypes parametersTypes = null)
+        {
+            byte[] bytes = Encoding.UTF8.GetBytes(jsonString);
```
joaompneves, line 58:
```suggestion
            var bytes = Encoding.UTF8.GetBytes(jsonString);
```

CefGlue.Common.Shared/Serialization/Deserializer.cs
```diff
@@ -0,0 +1,275 @@
...
+                switch (reader.TokenType)
+                {
+                    case JsonTokenType.Null:
+                        currentState.SetValue(null);
+                        break;
+
+                    case JsonTokenType.Number:
+                        currentState.SetValue(
+                            reader.GetNumber(currentState.ObjectTypeInfo.GetPropertyType(currentState.PropertyName))
+                            );
```
joaompneves, line 87: it makes more sense to use like this
```suggestion
                        currentState.SetValue(reader.GetNumber(currentState.GetPropertyType());
```

CefGlue.Common.Shared/Serialization/Deserializer.cs
```diff
@@ -0,0 +1,275 @@
...
+            object obj;
+            IDeserializerState newState = null;
+            JsonTypeInfo newTypeInfo = null;
+
+            tempReader.ReadToken(JsonTokenType.StartObject);
+
+            if (tempReader.TokenType == JsonTokenType.EndObject)
+            {
+                // empty object
+                obj = new object();
```
joaompneves, line 157: hum.. you would be pushing null to the current state, is that ok? have a test that falls into this case?

CefGlue.Common.Shared/Serialization/Deserializer.cs
```diff
@@ -0,0 +1,275 @@
...
+            }
+            else
+            {
+                // peek first field name
+                var propName = tempReader.ReadPropertyName();
+                IDeserializerState currentState;
+                switch (propName)
+                {
+                    case JsonAttributes.Ref:
```
joaompneves, line 166: missing here a comment with a json snippet of the expected json that will be read

CefGlue.Common.Shared/Serialization/Deserializer.cs
```diff
@@ -0,0 +1,275 @@
...
+                        reader.ReadToken(JsonTokenType.StartObject);
+                        reader.ReadPropertyName();  // skip the $ref
+                        var refId = reader.GetString();
+                        var refInfo = referencesMap[refId];
+                        obj = refInfo.Object;
+                        newTypeInfo = refInfo.TypeInfo;
+                        newState = CreateNewDeserializerState(obj, newTypeInfo);
```
joaompneves, line 173:
```suggestion
                        newState = CreateNewDeserializerState(refInfo.Object, refInfo.TypeInfo);
```

CefGlue.Common.Shared/Serialization/Deserializer.cs
```diff
@@ -0,0 +1,275 @@
...
+                        break;
+
+                    case JsonAttributes.Id:
+                        reader.ReadToken(JsonTokenType.StartObject);
+                        reader.ReadPropertyName(); // skip the $id
+                        var id = tempReader.ReadString();
+                        propName = tempReader.ReadPropertyName();
+                        if (propName == JsonAttributes.Values)
```
joaompneves, line 181: avoid reusing variables
```suggestion
                        if (tempReader.ReadPropertyName() == JsonAttributes.Values)
```

CefGlue.Common.Shared/Serialization/Deserializer.cs
```diff
@@ -0,0 +1,275 @@
...
+
+            return root.First();
+        }
+
+        private static object ReadComplexObject(ref Utf8JsonReader reader, Stack<IDeserializerState> state, bool isRootObjectHolder, ParametersTypes parametersTypes, IDictionary<string, ReferenceInfo> referencesMap)
+        {
+            var tempReader = reader;
+            object obj;
+            IDeserializerState newState = null;
+            JsonTypeInfo newTypeInfo = null;
```
joaompneves, line 150: only the newState is necessary... the other vars are for local scopes... declare them inplace where needed

CefGlue.Common.Shared/Serialization/Deserializer.cs
```diff
@@ -0,0 +1,275 @@
...
+
+        /// <summary>
+        /// Deserializes the passed jsonString argument to an instance of the TargetType.
+        /// </summary>
+        /// <typeparam name="TargetType"></typeparam>
+        /// <param name="jsonString"></param>
+        /// <param name="parametersTypes">When deserializing an argumentsArray string, it's possible to specify the specific typeInfo of each of the arguments.</param>
+        /// <returns></returns>
+        /// <exception cref="InvalidOperationException"></exception>
+        public static TargetType Deserialize<TargetType>(string jsonString, ParametersTypes parametersTypes = null)
```
joaompneves, line 38: work with type info abstraction and remove parameter types all over

CefGlue.Common.Shared/Serialization/Deserializer.cs
```diff
@@ -0,0 +1,275 @@
...
+                        break;
+                }
+            }
+
+            state.Push(newState);
+
+            return obj;
+        }
+
+        private static IDeserializerState CreateNewDeserializerState(
```
joaompneves, line 217: hum.. all these CreateNewDeserializerState seem a little complicated... I believe theres room for simplification

joaompneves (commented):
not finished reviewing but there's already alot of stuff to address

CefGlue.BrowserProcess/ObjectBinding/CefGlueGlobalScript.js
```diff
@@ -0,0 +1,152 @@
...
+                if (value === null) {
+                    return null;
+                }
+                const valueType = typeof value;
+                // case the property is of type Date, the argument 'value' will be of type 'string',
+                // in this case, we need to return the date as a string, but prefixed with a marker,
+                // so it can be properly deserialized
+                if (valueType === "string" && Reflect.get(this, key) instanceof Date) {
+                    return "$DateTimeMarker$" + value;
+                }
```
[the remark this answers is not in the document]
brunocunhasilva: Please check the changes to this commit - I have doubts about the identification of a binary value (checking if it's an instance of Uint8Array).
image

CefGlue.Common.Shared/Serialization/JsonReaderExtensions.cs
```diff
@@ -0,0 +1,128 @@
...
+                    return reader.GetUInt64();
+                case TypeCode.Single:
+                    return reader.GetSingle();
+                case TypeCode.Double:
+                    return reader.GetDouble();
+                case TypeCode.Decimal:
+                    return reader.GetDecimal();
+                default:
+                    // e.g. convert to the object type used in ExpandoObjects
+                    return Convert.ChangeType(reader.GetDouble(), typeToConvert);
```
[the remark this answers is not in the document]
brunocunhasilva: Yes - for example, check the `JavascriptEvaluationTests.DynamicObjectReturn` test.
That's why I left the comment

CefGlue.Common.Shared/Serialization/Deserializer.cs
brunocunhasilva: There's seems to exist a limitation in the JsonSerializer that throws an exception when serializing objects deeper than 1000 levels - this was causing an error in some tests I did, which ended up with me setting a lower value. But I guess our code should not be concerned by this internal limitation of the JsonSerializer, one that can be changed in future versions.
image


I'll set this maxDepth to be `int.MaxValue`

CefGlue.Common.Shared/Serialization/Deserializer.cs
brunocunhasilva: What type info abstraction are you referring to? (When deserializing an argumentsArray string, it's TargetType will be `object[]`, but we need to deserialize each one of its elements to the correct argument type)

Can you be a bit more specific?

CefGlue.Common.Shared/Serialization/Deserializer.cs
brunocunhasilva: Only the `newTypeInfo` variable can be moved, the other variables are used within the current scope.

CefGlue.Common.Shared/Serialization/Deserializer.cs
brunocunhasilva: Nice catch. I had forgotten to create a test that validated plain objects (e.g. `new object()`).
I created one and fixed this part of the code.

CefGlue.Common.Shared/Serialization/Deserializer.cs
brunocunhasilva: Refactored the code a bit.

CefGlue.BrowserProcess/Handlers/RenderProcessHandler.cs
```diff
@@ -23,7 +21,7 @@ public RenderProcessHandler()
         protected override void OnWebKitInitialized()
         {
             base.OnWebKitInitialized();
-            _javascriptExecutionEngine = new JavascriptExecutionEngineRenderSide(_messageDispatcher);
+            JavascriptHelper.RegisterJavascriptExecution(_messageDispatcher);
```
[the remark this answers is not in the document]
brunocunhasilva: I don't understand what you mean

CefGlue.Common.Shared/Serialization/Deserializer.cs
brunocunhasilva: Check the changes made in [this commit](https://github.com/OutSystems/CefGlue/pull/55/commits/f3eacc2d3b7b4ec79b54ba360b0bcba2d08ccfb0).
The code is not as resource efficient as it was, but it complies with the responsibilities isolation, namely in the Deserializer not knowing what ParameterTypes are and their specifics.

CefGlue.Common/ObjectBinding/NativeMethod.cs
```diff
@@ -72,30 +81,98 @@ private object ExecuteMethod(object targetObj, object[] args)
...
+                ValidateMandatoryArguments(nativeArgs);
+                return nativeArgs;
             }
 
+            var originalArgs = (object[])argsAsObject;
+            
+            ValidateMandatoryArguments(originalArgs);
+
             var argIndex = 0;
             var convertedArgs = new List<object>(_mandatoryParameters.Length + 1);
```
joaompneves, line 99: we should create and array here right away with the final size... to avoid and extra allocation

CefGlue.Common/ObjectBinding/NativeMethod.cs
```diff
@@ -72,30 +81,98 @@ private object ExecuteMethod(object targetObj, object[] args)
...
-                convertedArgs.Add(JavascriptToNativeTypeConverter.ConvertToNative(optionalArgs, _optionalParameter.ParameterType));
+                // the optionalParameterType is always a ParamArray of some type (eg int[]), so we need the ElementType
+                var optionalArgsArray = ExtractArray(originalArgs, argIndex, _optionalParameter.ParameterType.GetElementType());
+                convertedArgs.Add(optionalArgsArray);
             }
 
             return convertedArgs.ToArray();
         }
 
+        private object[] ConvertToNative(string args)
```
joaompneves, line 116: InnerConvertArguments

CefGlue.Common.Shared/Serialization/Deserializer.cs
```diff
@@ -0,0 +1,285 @@
...
+        private static object InnerDeserialize(string jsonString, Type[] targetTypes)
+        {
+            var bytes = Encoding.UTF8.GetBytes(jsonString);
+            var reader = new Utf8JsonReader(bytes, new JsonReaderOptions() { MaxDepth = DeserializerMaxDepth });
+
+            reader.Read();
+            return Deserialize(ref reader, targetTypes);
+        }
+
+        private static object Deserialize(ref Utf8JsonReader reader, Type[] targetTypes)
```
joaompneves, line 85:
```suggestion
        private static object InnerDeserialize(ref Utf8JsonReader reader, Type[] targetTypes)
```

CefGlue.Common.Shared/Serialization/Deserializer.cs
```diff
@@ -0,0 +1,285 @@
...
+        {
+            var state = new Stack<IDeserializerState>();
+            var root = new object[1];
+            var referencesMap = new Dictionary<string, ReferenceInfo>();
+            IDeserializerState previousStateIfEndArray = null;
+
+            state.Push(new ArrayDeserializerState(root, targetTypes[0]));
```
joaompneves, line 92: why just the first targetType?

CefGlue.Common.Shared/Serialization/Deserializer.cs
```diff
@@ -0,0 +1,285 @@
...
+                    case JsonTokenType.True:
+                        currentState.SetValue(reader.GetBoolean());
+                        break;
+
+                    case JsonTokenType.PropertyName:
+                        currentState.PropertyName = reader.GetString();
+                        break;
+
+                    case JsonTokenType.StartArray:
+                        var isRootObjectHolder = currentState.ObjectHolder == root;
```
joaompneves, line 122: why is root a special case? something is not right

CefGlue.Common/ObjectBinding/NativeMethod.cs
```diff
@@ -72,30 +81,98 @@ private object ExecuteMethod(object targetObj, object[] args)
...
+            }
+
+            var targetTypes = _mandatoryParameters.Select(p => p.ParameterType);
+            Type optionalParameterElementType = null;
+            
+            if (_optionalParameter != null)
+            {
+                // the optionalParameterType is always a ParamArray of some type (eg int[]), so we need the ElementType
+                optionalParameterElementType = _optionalParameter.ParameterType.GetElementType();
+                targetTypes = targetTypes.Append(optionalParameterElementType);
```
joaompneves, line 130: why the element type and not the the paramType?

Deserializer should receive information that you want an array of X type, that way the hack below wouldn't also be needed

CefGlue.BrowserProcess/ObjectBinding/CefGlueGlobalScript.js
```diff
@@ -0,0 +1,165 @@
...
+            const byteCharacters = atob(value);
+            const byteArray = new Array(byteCharacters.length);
+            for (let i = 0; i < byteCharacters.length; i++) {
+                byteArray[i] = byteCharacters.charCodeAt(i);
+            }
+            return byteArray;
+        }
+        function convertBinaryToBase64(byteArray) {
+            return btoa(String.fromCharCode(...byteArray));
+        }
```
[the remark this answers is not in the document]
brunocunhasilva: Not quite, because the test `JavascriptEvaluationTests.BinaryReturn` if I use the `btoa(byteArray)` here

CefGlue.Common/ObjectBinding/NativeMethod.cs
brunocunhasilva: The problem lies with the jsonString itelf, which is not an array for the optional parameters.
Arguments are in the form of an array, but the optional ones are not a sub-array.
e.g. this is an example of what the json string looks like:
"[arg1, arg2, optionalArg1, optionalArg2]"

To work as you are asking, the string would have to be in the form of:
"[arg1, arg2, [optionalArg1, optionalArg2]]"

CefGlue.Common.Shared/Serialization/Deserializer.cs
brunocunhasilva: It was a way of distinguishing between the root `ArrayDeserializerState` and the `ArrayDeserializerState` that is used for the "arguments" array (in https://github.com/OutSystems/CefGlue/blob/53baca0223c77d6f7f91a4c99073021db31e054d/CefGlue.Common.Shared/Serialization/Deserializer.cs#L249)

Nevertheless, I ended up refactoring the code and removing the need to check for the isRootObjectHolder

CefGlue.Common.Shared/Serialization/State/JsonTypeInfo.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace Xilium.CefGlue.Common.Shared.Serialization.State
+{
+    internal record JsonTypeInfo
+    {
+        public JsonTypeInfo(Type type)
+        {
+            ObjectType = type;
+        }
+        public Type ObjectType { get; set; }
+        public Dictionary<string, TypeMemberInfo> TypeMembers { get; } = new Dictionary<string, TypeMemberInfo>();
```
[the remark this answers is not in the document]
brunocunhasilva: If it becomes a IReaOnlyDictionary, than the Add of elements can't be done in `JsonTypeInfoCache`, as it is now. And an `AddTypeMember` method must be created inside of the `JsonTypeInfo` class (switched from Record to class considering it'll have some login inside)

CefGlue.Common.Shared/Serialization/State/ArrayDeserializerState.cs
```diff
@@ -0,0 +1,60 @@
...
+        internal static ArrayDeserializerState Create(Utf8JsonReader reader, JsonTypeInfo objectTypeInfo, string propertyName)
+        {
+            var newArray = CreateArray(reader, objectTypeInfo, propertyName, out var arrayElementType);
+            return new ArrayDeserializerState(newArray, arrayElementType);
+        }
+
+        public override JsonTypeInfo ObjectTypeInfo =>
+            JsonTypeInfoCache.GetOrAddTypeInfo(
+                _targetTypes[_arrayIndex < _targetTypes.Length ? _arrayIndex : _targetTypes.Length - 1]
+                );
```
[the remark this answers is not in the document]
brunocunhasilva: What syntax are you referring to? (the way the method is split, the usage of a ternary expression, not using the explicit get)?

CefGlue.Common.Shared/Serialization/State/JsonTypeInfo.cs
```diff
@@ -0,0 +1,51 @@
...
+            foreach (var field in fields)
+            {
+                _typeMembers.Add(field.Name, new TypeMemberInfo(field.FieldType, (obj, value) => field.SetValue(obj, value)));
+            }
+
+            MethodInfo addMethod;
+            if (ObjectType.IsCollection() && (addMethod = ObjectType.GetMethod("Add")) != null)
+            {
+                CollectionAddMethod = new TypeMethodInfo((obj, value) => addMethod.Invoke(obj, value));
+            }
```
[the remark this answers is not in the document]
brunocunhasilva: refactored

PR #61, opened 2023-01-23
Removing slow opening and runtime errors

osesantos:
These changes will remove some errors when starting the application. The Browser process was looking for the `libGLESv2.dylib` and was not able to find it because the file was being added to the root Build folder.
Here we will add all the cef files required by the browser process to run.


CefGlue.Common/CefGlue.Common.csproj
```diff
@@ -47,16 +47,15 @@
     <Target DependsOnTargets="ResolveReferences" Name="CopyProjectReferencesToPackage">
         <ItemGroup>
             <BuildOutputInPackage Include="@(ReferenceCopyLocalPaths-&gt;WithMetadataValue('ReferenceSourceTarget', 'ProjectReference'))" />
-            <BuildOutputInPackage Condition="'$(TargetFramework)' == 'net472'" Include="..\CefGlue.BrowserProcess\bin\$(Configuration)\net472\Xilium.CefGlue.BrowserProcess.*" />
 
-            <TfmSpecificPackageFile Include="..\CefGlue.BrowserProcess\bin\$(Configuration)\net5.0\win-x64\publish\**\*">
+            <TfmSpecificPackageFile Include="..\CefGlue.BrowserProcess\bin\$(Configuration)\net6.0\win-x64\publish\**\*">
```
joaompneves, line 51: create a property to specify the net version

CefGlue.Common/build/CefGlue.Common.targets
```diff
@@ -17,6 +17,15 @@
             <PublishState>Included</PublishState>
         </None>
     </ItemGroup>
+    <ItemGroup Condition="'$(CefGlueTargetPlatform)' == 'osx'">
```
joaompneves, line 20: merge the 2 sections under the same condition

CefGlue.Common/build/CefGlue.Common.targets
```diff
@@ -17,6 +17,15 @@
...
+        <None Include="@(CefRedistOSX64)">
+            <Visible>false</Visible>
+            <Link>$(OutputDirectory)\CefGlueBrowserProcess\%(RecursiveDir)%(FileName)%(Extension)</Link>
```
joaompneves, line 23: whats the folder output ... will we have repeated files?

osesantos: we will have some repeated files, and these are an example of them.
image

However, because these are 2 independent processes and both processes now require these files to run, I don't see a problem with having them in both dirs.

joaompneves: I'm trying to understand what changed, because the process architecture didn't change... so why is this now requiring this.... also whats the size of the highlight files combined?

CefGlue.Common/CefGlue.Common.csproj
```diff
@@ -47,16 +47,15 @@
     <Target DependsOnTargets="ResolveReferences" Name="CopyProjectReferencesToPackage">
         <ItemGroup>
             <BuildOutputInPackage Include="@(ReferenceCopyLocalPaths-&gt;WithMetadataValue('ReferenceSourceTarget', 'ProjectReference'))" />
-            <BuildOutputInPackage Condition="'$(TargetFramework)' == 'net472'" Include="..\CefGlue.BrowserProcess\bin\$(Configuration)\net472\Xilium.CefGlue.BrowserProcess.*" />
 
-            <TfmSpecificPackageFile Include="..\CefGlue.BrowserProcess\bin\$(Configuration)\net5.0\win-x64\publish\**\*">
+            <TfmSpecificPackageFile Include="..\CefGlue.BrowserProcess\bin\$(Configuration)\$(DotnetVersion)\win-x64\publish\**\*">
```
joaompneves, line 51: TargetFramework

CefGlue.Common/CefGlue.Common.csproj
```diff
@@ -47,16 +47,15 @@
...
                 <PackagePath>bin\win</PackagePath>
             </TfmSpecificPackageFile>
 
-            <TfmSpecificPackageFile Include="..\CefGlue.BrowserProcess\bin\$(Configuration)\net5.0\osx-x64\publish\**\*">
+            <TfmSpecificPackageFile Include="..\CefGlue.BrowserProcess\bin\$(Configuration)\$(DotnetVersion)\osx-x64\publish\**\*">
```
joaompneves, line 55: TargetFramework

CefGlue.Common/CefGlue.Common.csproj
osesantos: since we have 2 target frameworks (netstandard2.0 and net6) I prefer having something like DotnetVersion for .ne6 and DotnetStandardVersion for netstandard2.0

CefGlue.Common/CefGlue.Common.csproj
osesantos: since we have 2 target frameworks (netstandard2.0 and net6) I prefer having something like DotnetVersion for .ne6 and DotnetStandardVersion for netstandard2.0

joaompneves: yeah ... but what I meant here is to use the variable from this project.. that was set above rather than the global one... it makes more sense

joaompneves: nvm.. this project is .net standard

CefGlue.Common/build/CefGlue.Common.targets
osesantos: also don't know... what caused this, I and @cangosta talked about it and looked a bit but because there are some many changes on the cef side it's hard to tell which one was.
about the size, the nuget has around 200MB, so all these files will be in both folders
image

joaompneves: this needs to be investigated because it will simply double the size of cefglue on disk (200 mbs -> 400 ... round numbers)

osesantos: Created a new target that will move the required files instead of copy

CefGlue.Common/build/CefGlue.Common.targets
```diff
@@ -26,4 +26,14 @@
...
     </ItemGroup>
-</Project>
\ No newline at end of file
+
+    <ItemGroup>
+        <FilesToMove Include="$(OutDir)\libEGL.dylib"/>
+        <FilesToMove Include="$(OutDir)\libGLESv2.dylib"/>
+        <FilesToMove Include="$(OutDir)\libvk_swiftshader.dylib"/>
+    </ItemGroup>
+    <Target Name="MoveCeflib" AfterTargets="Build">
```
joaompneves, line 35:
```suggestion
    <Target Name="MoveCeflib" AfterTargets="Build">
            <ItemGroup>
                <FilesToMove Include="$(OutDir)\libEGL.dylib"/>
                <FilesToMove Include="$(OutDir)\libGLESv2.dylib"/>
                <FilesToMove Include="$(OutDir)\libvk_swiftshader.dylib"/>
           </ItemGroup>
```

CefGlue.Common/build/CefGlue.Common.targets
```diff
@@ -26,4 +26,14 @@
...
+        <Message Text="Moving Files @(FilesToMove)[$(ProjectFileName)]"/>
+        <Move SourceFiles="@(FilesToMove)" DestinationFolder="$(OutDir)\CefGlueBrowserProcess\"/>
```
joaompneves, line 37: create a variable for CefGlueBrowserProcess

CefGlue.Common/build/CefGlue.Common.targets
```diff
@@ -20,10 +24,20 @@
...
             <CopyToPublishDirectory>PreserveNewest</CopyToPublishDirectory>
             <PublishState>Included</PublishState>
         </None>
     </ItemGroup>
-</Project>
\ No newline at end of file
+
+    <Target Name="MoveCeflib" AfterTargets="Build">
+        <ItemGroup>
+            <FilesToMove Include="$(OutDir)\libEGL.dylib"/>
```
joaompneves, line 36: on windows... does it complain with these file moves during build?

osesantos: no, this was something we confirm that is only happening on mac

joaompneves: yeah.. but the moves also happen when building on windows, because there's no condition here

osesantos: true, with the change from copy to move and a new target, forgot to also move the condition. Nice catch

joaompneves (approved):

PR #74, opened 2023-02-17
Add RequestContext accessor

Wiks00:
Hello.
Would like to introduce functionality to add a request context to the browser creation process.

### DIFF
--- a/CefGlue.Avalonia/AvaloniaCefBrowser.cs
+++ b/CefGlue.Avalonia/AvaloniaCefBrowser.cs
@@ -1,4 +1,5 @@
-﻿using Xilium.CefGlue.Avalonia.Platform;
+﻿using System;
+using Xilium.CefGlue.Avalonia.Platform;
 using Xilium.CefGlue.Common;
 using Xilium.CefGlue.Common.Platform;
 
@@ -17,10 +18,8 @@ namespace Xilium.CefGlue.Avalonia
             }
         }
 
-        public AvaloniaCefBrowser() : base(null) { }
-
-        public AvaloniaCefBrowser(CefRequestContextSettings cefRequestContextSettings, CefRequestContextHandler cefRequestContextHandler = null)
-            : base(cefRequestContextSettings, cefRequestContextHandler)
+        public AvaloniaCefBrowser(Func<CefRequestContext> cefRequestContextFactory = null)
+            : base(cefRequestContextFactory)
         { }
 
         internal override Common.Platform.IControl CreateControl()
--- a/CefGlue.Avalonia/AvaloniaCefBrowser.cs
+++ b/CefGlue.Avalonia/AvaloniaCefBrowser.cs
@@ -17,6 +17,18 @@ namespace Xilium.CefGlue.Avalonia
             }
         }
 
+        public AvaloniaCefBrowser()
+            : base(null)
+        { }
+
+        public AvaloniaCefBrowser(CefRequestContextSettings cefRequestContextSettings, CefRequestContextHandler cefRequestContextHandler = null)
+            : base(cefRequestContextSettings, cefRequestContextHandler)
+        { }
+
+        public AvaloniaCefBrowser(CefRequestContext cefRequestContext, CefRequestContextHandler cefRequestContextHandler = null)
+            : base(cefRequestContext, cefRequestContextHandler)
+        { }
+
         internal override Common.Platform.IControl CreateControl()
         {
             return new AvaloniaControl(this, VisualChildren);
--- a/CefGlue.Avalonia/AvaloniaCefBrowser.cs
+++ b/CefGlue.Avalonia/AvaloniaCefBrowser.cs
@@ -25,10 +25,6 @@ namespace Xilium.CefGlue.Avalonia
             : base(cefRequestContextSettings, cefRequestContextHandler)
         { }
 
-        public AvaloniaCefBrowser(CefRequestContext cefRequestContext, CefRequestContextHandler cefRequestContextHandler = null)
-            : base(cefRequestContext, cefRequestContextHandler)
-        { }
-
         internal override Common.Platform.IControl CreateControl()
         {
             return new AvaloniaControl(this, VisualChildren);
--- a/CefGlue.Avalonia/AvaloniaCefBrowser.cs
+++ b/CefGlue.Avalonia/AvaloniaCefBrowser.cs
@@ -17,9 +17,7 @@ namespace Xilium.CefGlue.Avalonia
             }
         }
 
-        public AvaloniaCefBrowser()
-            : base(null)
-        { }
+        public AvaloniaCefBrowser() : base(null) { }
 
         public AvaloniaCefBrowser(CefRequestContextSettings cefRequestContextSettings, CefRequestContextHandler cefRequestContextHandler = null)
             : base(cefRequestContextSettings, cefRequestContextHandler)
--- a/CefGlue.Common/BaseCefBrowser.cs
+++ b/CefGlue.Common/BaseCefBrowser.cs
@@ -30,10 +30,6 @@ namespace Xilium.CefGlue.Common
             : this(CefRequestContext.CreateContext(cefRequestContextSettings, cefRequestContextHandler))
         { }
 
-        public BaseCefBrowser(CefRequestContext cefRequestContext, CefRequestContextHandler cefRequestContextHandler = null)
-            : this(CefRequestContext.CreateContext(cefRequestContext, cefRequestContextHandler))
-        { }
-
         public BaseCefBrowser(CefRequestContext cefRequestContext = null)
         {
 #if HAS_NLOG
--- a/CefGlue.Common/BaseCefBrowser.cs
+++ b/CefGlue.Common/BaseCefBrowser.cs
@@ -18,13 +18,24 @@ namespace Xilium.CefGlue.Common
 
         #region Disposable
 
-        public BaseCefBrowser()
+        static BaseCefBrowser()
         {
             if (!CefRuntimeLoader.IsLoaded)
             {
                 CefRuntimeLoader.Load();
             }
+        }
 
+        public BaseCefBrowser(CefRequestContextSettings cefRequestContextSettings, CefRequestContextHandler cefRequestContextHandler = null)
+            : this(CefRequestContext.CreateContext(cefRequestContextSettings, cefRequestContextHandler))
+        { }
+
+        public BaseCefBrowser(CefRequestContext cefRequestContext, CefRequestContextHandler cefRequestContextHandler = null)
+            : this(CefRequestContext.CreateContext(cefRequestContext, cefRequestContextHandler))
+        { }
+
+        public BaseCefBrowser(CefRequestContext cefRequestContext = null)
+        {
 #if HAS_NLOG
             _logger = new Logger(nameof(BaseCefBrowser));
 #else
@@ -33,11 +44,11 @@ namespace Xilium.CefGlue.Common
 
             if (CefRuntimeLoader.IsOSREnabled)
             {
-                _adapter = new CommonOffscreenBrowserAdapter(this, nameof(BaseCefBrowser), CreateOffScreenControlHost(), CreatePopupHost(), _logger);
+                _adapter = new CommonOffscreenBrowserAdapter(this, nameof(BaseCefBrowser), CreateOffScreenControlHost(), CreatePopupHost(), _logger, cefRequestContext);
             } 
             else
             {
-                _adapter = new CommonBrowserAdapter(this, nameof(BaseCefBrowser), CreateControl(), _logger);
+                _adapter = new CommonBrowserAdapter(this, nameof(BaseCefBrowser), CreateControl(), _logger, cefRequestContext);
             }
         }
 
@@ -239,20 +250,9 @@ namespace Xilium.CefGlue.Common
         public double ZoomLevel { get => _adapter.ZoomLevel; set => _adapter.ZoomLevel = value; }
 
         /// <summary>
-        /// Get or set <see cref="CefRequestContext"/>
+        /// Get <see cref="CefRequestContext"/>
         /// </summary>
-        public CefRequestContext RequestContext 
-        {
-            get
-            {
-                return _adapter.RequestContext;
-            }
-
-            set
-            {
-                _adapter.RequestContext = value;
-            }
-        }
+        public CefRequestContext RequestContext { get => _adapter.RequestContext; }
 
         /// <summary>
         /// The undelying cef browser instance. Can be used for advanced functionality.
--- a/CefGlue.Common/BaseCefBrowser.cs
+++ b/CefGlue.Common/BaseCefBrowser.cs
@@ -33,11 +33,11 @@ namespace Xilium.CefGlue.Common
 
             if (CefRuntimeLoader.IsOSREnabled)
             {
-                _adapter = new CommonOffscreenBrowserAdapter(this, nameof(BaseCefBrowser), CreateOffScreenControlHost(), CreatePopupHost(), _logger, cefRequestContextFactory());
+                _adapter = new CommonOffscreenBrowserAdapter(this, nameof(BaseCefBrowser), CreateOffScreenControlHost(), CreatePopupHost(), _logger, cefRequestContextFactory?.Invoke());
             } 
             else
             {
-                _adapter = new CommonBrowserAdapter(this, nameof(BaseCefBrowser), CreateControl(), _logger, cefRequestContextFactory());
+                _adapter = new CommonBrowserAdapter(this, nameof(BaseCefBrowser), CreateControl(), _logger, cefRequestContextFactory?.Invoke());
             }
         }
 
--- a/CefGlue.Common/BaseCefBrowser.cs
+++ b/CefGlue.Common/BaseCefBrowser.cs
@@ -18,20 +18,13 @@ namespace Xilium.CefGlue.Common
 
         #region Disposable
 
-        static BaseCefBrowser()
+        public BaseCefBrowser(Func<CefRequestContext> cefRequestContextFactory = null)
         {
             if (!CefRuntimeLoader.IsLoaded)
             {
                 CefRuntimeLoader.Load();
             }
-        }
 
-        public BaseCefBrowser(CefRequestContextSettings cefRequestContextSettings, CefRequestContextHandler cefRequestContextHandler = null)
-            : this(CefRequestContext.CreateContext(cefRequestContextSettings, cefRequestContextHandler))
-        { }
-
-        public BaseCefBrowser(CefRequestContext cefRequestContext = null)
-        {
 #if HAS_NLOG
             _logger = new Logger(nameof(BaseCefBrowser));
 #else
@@ -40,11 +33,11 @@ namespace Xilium.CefGlue.Common
 
             if (CefRuntimeLoader.IsOSREnabled)
             {
-                _adapter = new CommonOffscreenBrowserAdapter(this, nameof(BaseCefBrowser), CreateOffScreenControlHost(), CreatePopupHost(), _logger, cefRequestContext);
+                _adapter = new CommonOffscreenBrowserAdapter(this, nameof(BaseCefBrowser), CreateOffScreenControlHost(), CreatePopupHost(), _logger, cefRequestContextFactory());
             } 
             else
             {
-                _adapter = new CommonBrowserAdapter(this, nameof(BaseCefBrowser), CreateControl(), _logger, cefRequestContext);
+                _adapter = new CommonBrowserAdapter(this, nameof(BaseCefBrowser), CreateControl(), _logger, cefRequestContextFactory());
             }
         }
 
--- a/CefGlue.Common/BaseCefBrowser.cs
+++ b/CefGlue.Common/BaseCefBrowser.cs
@@ -239,6 +239,22 @@ namespace Xilium.CefGlue.Common
         public double ZoomLevel { get => _adapter.ZoomLevel; set => _adapter.ZoomLevel = value; }
 
         /// <summary>
+        /// Get or set <see cref="CefRequestContext"/>
+        /// </summary>
+        public CefRequestContext RequestContext 
+        {
+            get
+            {
+                return _adapter.RequestContext;
+            }
+
+            set
+            {
+                _adapter.RequestContext = value;
+            }
+        }
+
+        /// <summary>
         /// The undelying cef browser instance. Can be used for advanced functionality.
         /// </summary>
         protected CefBrowser UnderlyingBrowser => _adapter.Browser;
--- a/CefGlue.Common/CommonBrowserAdapter.cs
+++ b/CefGlue.Common/CommonBrowserAdapter.cs
@@ -33,13 +33,14 @@ namespace Xilium.CefGlue.Common
 
         private object _disposeLock = new object();
 
-        public CommonBrowserAdapter(object eventsEmitter, string name, IControl control, ILogger logger)
+        public CommonBrowserAdapter(object eventsEmitter, string name, IControl control, ILogger logger, CefRequestContext cefRequestContext = null)
         {
             _eventsEmitter = eventsEmitter;
             _name = name;
             _logger = logger;
 
             Control = control;
+            RequestContext = cefRequestContext;
 
             control.GotFocus += HandleGotFocus;
             control.SizeChanged += HandleControlSizeChanged;
@@ -118,7 +119,7 @@ namespace Xilium.CefGlue.Common
 
         public event AsyncUnhandledExceptionEventHandler UnhandledException;
 
-        public CefRequestContext RequestContext { get; set; }
+        public CefRequestContext RequestContext { get; init; }
 
         public string Address { get => _browser?.GetMainFrame().Url ?? _initialUrl; set => NavigateTo(value); }
 
@@ -195,7 +196,7 @@ namespace Xilium.CefGlue.Common
 
         public bool CanGoForward()
         {
-            return _browser?.CanGoForward ?? false;    
+            return _browser?.CanGoForward ?? false;
         }
 
         public void GoForward()
@@ -443,7 +444,7 @@ namespace Xilium.CefGlue.Common
                     _browser?.GetMainFrame()?.LoadUrl(_initialUrl);
                     _initialUrl = "";
                 }
-                
+
                 Initialized?.Invoke();
             });
         }
@@ -458,7 +459,7 @@ namespace Xilium.CefGlue.Common
             if (browser.IsPopup)
             {
                 // popup such as devtools, let it close its window
-                return false; 
+                return false;
             }
 
             Control.DestroyRender();
@@ -593,7 +594,7 @@ namespace Xilium.CefGlue.Common
         {
             Control.CloseContextMenu();
         }
-       
+
         void ICefBrowserHost.HandleException(Exception exception)
         {
             HandleException("Unknown", exception);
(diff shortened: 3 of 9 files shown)


Wiks00:
> Hello @Wiks00 👋
> 
> Thanks for your Pull Request. It looks like this may be your first contribution to an OutSystems Open Source project. Before we look into your contribution, we need you to accept the 📝 [OutSystems Contributors License Agreement](https://outsystems-public-dl.s3-eu-west-1.amazonaws.com/opensource/cla/OutSystems-Contributor-License-Agreement-2020-03-23.pdf).
> 
> After you read and accept the terms of the agreement, reply to this message saying `I read, understood and accept the OutSystems CLA`.

I read, understood and accept the OutSystems CLA

Wiks00:
@joaompneves, @dmitry-azaraev Could anyone take a look ?
Thanks.

dmitry-azaraev:
> @dmitry-azaraev Could anyone take a look ? Thanks.

I'm not an owner of this repository, nor functionality of changed codepaths. CefGlue lives only in https://gitlab.com/xiliumhq/chromiumembedded/cefglue and actually i'm care only about `Xilium.CefGlue` assembly. Everything else is out of my scope.

CefGlue.Common/BaseCefBrowser.cs
```diff
@@ -238,6 +238,22 @@ protected virtual void Dispose(bool disposing)
         /// </summary>
         public double ZoomLevel { get => _adapter.ZoomLevel; set => _adapter.ZoomLevel = value; }
 
+        /// <summary>
+        /// Get or set <see cref="CefRequestContext"/>
+        /// </summary>
+        public CefRequestContext RequestContext 
+        {
+            get
```
dmitry-azaraev, line 246: code style: code around prefers more shorter/lambda forms.

CefGlue.Common/BaseCefBrowser.cs
```diff
@@ -238,6 +238,22 @@ protected virtual void Dispose(bool disposing)
...
+            {
+                return _adapter.RequestContext;
+            }
+
+            set
+            {
+                _adapter.RequestContext = value;
+            }
+        }
```
joaompneves, line 255:
```suggestion
        public CefRequestContext RequestContext { 
            get => _adapter.RequestContext;
            init => _adapter.RequestContext = value;
        }
```

CefGlue.Common/CommonBrowserAdapter.cs
```diff
@@ -118,6 +118,8 @@ public void Dispose(bool disposing)
 
         public event AsyncUnhandledExceptionEventHandler UnhandledException;
 
+        public CefRequestContext RequestContext { get; set; }
```
joaompneves, line 121:
```suggestion
        public CefRequestContext RequestContext { get; init; }
```

CefGlue.Demo.Avalonia/BrowserView.xaml.cs
```diff
@@ -27,6 +27,7 @@ public BrowserView()
             browser.RegisterJavascriptObject(new BindingTestClass(), "boundBeforeLoadObject");
             browser.LoadStart += OnBrowserLoadStart;
             browser.TitleChanged += OnBrowserTitleChanged;
+            browser.RequestContext = CefRequestContext.CreateContext(new CefRequestContextSettings(), null);
```
joaompneves, line 30: create a unit test for this functionality instead

joaompneves (changes requested):
Thank you for your contribution. I left some suggestions to improve the solution.

CefGlue.Common/CommonBrowserAdapter.cs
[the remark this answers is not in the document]
Wiks00: Good point. Moved initialization to ctor

Wiks00:
> > @dmitry-azaraev Could anyone take a look ? Thanks.
> 
> I'm not an owner of this repository, nor functionality of changed codepaths. CefGlue lives only in https://gitlab.com/xiliumhq/chromiumembedded/cefglue and actually i'm care only about `Xilium.CefGlue` assembly. Everything else is out of my scope.


Got it, sorry for the confusion. 
Anyway thanks for reply.

CefGlue.Common/BaseCefBrowser.cs
```diff
@@ -18,13 +18,20 @@ public abstract partial class BaseCefBrowser : IDisposable
...
         #region Disposable
 
-        public BaseCefBrowser()
+        static BaseCefBrowser()
         {
             if (!CefRuntimeLoader.IsLoaded)
             {
                 CefRuntimeLoader.Load();
             }
+        }
```
joaompneves, line 27: please revert this change... the browser should only be loaded upon the first browser instantiation, not when you use the browser class for the first time... stick with the minimal changes needed for the PR

CefGlue.Demo.Avalonia/BrowserView.xaml.cs
Wiks00: Done

CefGlue.Common/CommonBrowserAdapter.cs
```diff
@@ -118,6 +119,8 @@ public void Dispose(bool disposing)
 
         public event AsyncUnhandledExceptionEventHandler UnhandledException;
 
+        public CefRequestContext RequestContext { get; init; }
```
joaompneves, line 122:
```suggestion
        public CefRequestContext RequestContext { get; }
```

CefGlue.Common/BaseCefBrowser.cs
```diff
@@ -18,13 +18,20 @@ public abstract partial class BaseCefBrowser : IDisposable
...
+
+        public BaseCefBrowser(CefRequestContextSettings cefRequestContextSettings, CefRequestContextHandler cefRequestContextHandler = null)
+            : this(CefRequestContext.CreateContext(cefRequestContextSettings, cefRequestContextHandler))
+        { }
```
joaompneves, line 31: drop this overload... lets keep this simple

CefGlue.Avalonia/AvaloniaCefBrowser.cs
```diff
@@ -17,6 +17,14 @@ static AvaloniaCefBrowser()
             }
         }
 
+        public AvaloniaCefBrowser()
+            : base(null)
+        { }
+
+        public AvaloniaCefBrowser(CefRequestContextSettings cefRequestContextSettings, CefRequestContextHandler cefRequestContextHandler = null)
```
joaompneves, line 24:
```suggestion
        public AvaloniaCefBrowser(CefRequestContext cefRequestContext)
```

CefGlue.Avalonia/AvaloniaCefBrowser.cs
```diff
@@ -17,6 +17,14 @@ static AvaloniaCefBrowser()
             }
         }
 
+        public AvaloniaCefBrowser()
+            : base(null)
+        { }
```
joaompneves, line 22:
```suggestion
        public AvaloniaCefBrowser() : base(null) { }
```

CefGlue.Tests/TestBase.cs
```diff
@@ -103,14 +108,14 @@ protected virtual Task ExtraSetup()
         [TearDown] 
         protected void TearDown()
         {
-            browser.Dispose();
+            browser?.Dispose();
```
joaompneves, line 111: is this change needed?

CefGlue.Tests/TestBase.cs
```diff
@@ -103,14 +108,14 @@ protected virtual Task ExtraSetup()
...
         }
 
         [OneTimeTearDown]
         protected async Task OneTimeTearDown()
         {
             await Run(() => {
-                window.Close();
+                window?.Close();
```
joaompneves, line 118: needed?

CefGlue.Tests/RequestContext/CustomRequestContext.cs
```diff
@@ -0,0 +1,72 @@
...
+using Xilium.CefGlue;
+using Xilium.CefGlue.Avalonia;
+using Xilium.CefGlue.Common;
+using Xilium.CefGlue.Common.Shared;
+
+namespace CefGlue.Tests.RequestContext
+{ 
+    public class CustomRequestContext: TestBase
+    {
+        class CustomCefRequestContextHandler : CefRequestContextHandler
```
joaompneves, line 20:
```suggestion
        private class CustomCefRequestContextHandler : CefRequestContextHandler
```

CefGlue.Tests/RequestContext/CustomRequestContext.cs
```diff
@@ -0,0 +1,72 @@
...
+
+            await InternalSetup(browser);
+
+            Assert.False(Browser.RequestContext.IsGlobal);
+        }
+
+        [Test]
+        public async Task CefReqestContextApplied()
+        {
+            var cachePath = @"C:/path/to/cache";
```
joaompneves, line 45:
```suggestion
            const string CachePath = @"C:/path/to/cache";
```

CefGlue.Tests/RequestContext/CustomRequestContext.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        // skip base setup
+        protected override Task Setup()
+            => Task.CompletedTask;
+
+        [Test]
+        public async Task NotGlobalCefReqestContext()
+        {
+            var browser = () => new AvaloniaCefBrowser(new CefRequestContextSettings());
```
joaompneves, line 35:
```suggestion
            var browserFactory = () => new AvaloniaCefBrowser(new CefRequestContextSettings());
```

CefGlue.Tests/RequestContext/CustomRequestContext.cs
```diff
@@ -0,0 +1,72 @@
...
+
+            Assert.False(Browser.RequestContext.IsGlobal);
+        }
+
+        [Test]
+        public async Task CefReqestContextApplied()
+        {
+            var cachePath = @"C:/path/to/cache";
+
+            var browser = () => new AvaloniaCefBrowser(new CefRequestContextSettings()
```
joaompneves, line 47:
```suggestion
            var browserFactory = () => new AvaloniaCefBrowser(new CefRequestContextSettings()
```

CefGlue.Tests/RequestContext/CustomRequestContext.cs
```diff
@@ -0,0 +1,72 @@
...
+            var result = Browser.RequestContext.CachePath;
+
+            Assert.AreEqual(cachePath, result);
+        }
+
+        [Test]
+        public async Task CustomCefReqestContextHandlerApplied()
+        {
+            var customRequestContextHandler = new CustomCefRequestContextHandler();
+            var browser = () => new AvaloniaCefBrowser(new CefRequestContextSettings(), customRequestContextHandler);
```
joaompneves, line 63:
```suggestion
            var browserFactory = () => new AvaloniaCefBrowser(new CefRequestContextSettings(), customRequestContextHandler);
```

joaompneves (changes requested):

CefGlue.Common/BaseCefBrowser.cs
Wiks00: But does this piece of code represent browser initialization? In essence, the logic has not been changed. 
It is also impossible to create a RequestContext without first calling this block of code, which prevents the initialization of the RequestContext instance outside of this class.

CefGlue.Common/BaseCefBrowser.cs
Wiks00: Please see my comment about static ctor.

CefGlue.Tests/TestBase.cs
Wiks00: Yes, because if the test fails before running the `SetUp `code (in the case of RequestContext `InternalSetup`), this part of the test will crash.

CefGlue.Avalonia/AvaloniaCefBrowser.cs
Wiks00: Also closely related to browser initialization.

CefGlue.Tests/RequestContext/CustomRequestContext.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            await InternalSetup(browserFactory);
+
+            var result = Browser.RequestContext.CachePath;
+
+            Assert.AreEqual(cachePath, result);
+        }
+
+        [Test]
+        public async Task CustomCefReqestContextHandlerApplied()
```
joaompneves, line 59:
```suggestion
        public async Task CustomCefRequestContextHandlerApplied()
```

CefGlue.Tests/RequestContext/CustomRequestContext.cs
```diff
@@ -0,0 +1,71 @@
...
+        {
+            var browserFactory = () => new AvaloniaCefBrowser(new CefRequestContextSettings());
+
+            await InternalSetup(browserFactory);
+
+            Assert.False(Browser.RequestContext.IsGlobal);
+        }
+
+        [Test]
+        public async Task CefReqestContextApplied()
```
joaompneves, line 43:
```suggestion
        public async Task CefRequestContextApplied()
```

CefGlue.Tests/RequestContext/CustomRequestContext.cs
```diff
@@ -0,0 +1,71 @@
...
+                return null;
+            }
+        }
+
+        // skip base setup
+        protected override Task Setup()
+            => Task.CompletedTask;
+
+        [Test]
+        public async Task NotGlobalCefReqestContext()
```
joaompneves, line 33:
```suggestion
        public async Task NonGlobalCefRequestContext()
```

CefGlue.Common/BaseCefBrowser.cs
joaompneves: This piece of code basically loads cef dlls. I see what you're saying... but I'm also not very comfortable with this change because Load must be called on the main thread (typically the ui thread)... and with this change one can accidentally not call it on other threads

Wiks00: Based [on documentation](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/static-constructors#remarks) this is safe.
If the `BaseCefBrowser` instance is initialized on the UI thread, then the static ctor will run on the same thread.

Wiks00: Sorry, but I did not quite understand your point in the example.
The `typeof `operator takes the` System.Type` instance for the type and has nothing to do with `AvaloniaCefBrowser `and its base class. A static ctor initializes a class before the first instance is created or any static members _declared_ in that class (not its base classes) are referenced.

If your example runs on the UI thread, it is safe to say that no exceptions will be thrown on any UI framework

joaompneves: the example was not the best... but when you reference the class... example calling static method on that same class, the static ctor will run before (just once).

Wiks00: Okay, I think I get your point. Although there are no static members in `BaseCefBrowser`, which means that it is impossible to call it except to create an instance of the class, but it looks like "hidden knowledge".

Unfortunately, I do not see more "native" methods how to call initialization without code duplicates. Would this solution work?

```
        public BaseCefBrowser()
        {
            InitializeInternals();

            _logger = CreateBrowserLogger();

            _adapter = CreateBrowserAdapter();
        }

        public BaseCefBrowser(CefRequestContextSettings cefRequestContextSettings, CefRequestContextHandler cefRequestContextHandler = null)
        {
            InitializeInternals();

            _logger = CreateBrowserLogger();

            CefRequestContext cefRequestContext = CefRequestContext.CreateContext(cefRequestContextSettings, cefRequestContextHandler);

            _adapter = CreateBrowserAdapter(cefRequestContext);
        }

        private void InitializeInternals()
        {
            if (!CefRuntimeLoader.IsLoaded)
            {
                CefRuntimeLoader.Load();
            }
        }

        private ILogger CreateBrowserLogger()
        {

#if HAS_NLOG
            return new Logger(nameof(BaseCefBrowser));
#else
            return new NullLogger(nameof(BaseCefBrowser));
#endif
        }

        private CommonBrowserAdapter CreateBrowserAdapter(CefRequestContext cefRequestContext = null)
            => CefRuntimeLoader.IsOSREnabled ? 
                new CommonOffscreenBrowserAdapter(this, nameof(BaseCefBrowser), CreateOffScreenControlHost(), CreatePopupHost(), _logger, cefRequestContext) : 
                new CommonBrowserAdapter(this, nameof(BaseCefBrowser), CreateControl(), _logger, cefRequestContext);
```

joaompneves: it doesnt need that many changes, but that wont work either because in order to create a CefRequestContextHandler instance cef dlls need to be loaded. the way I see this working with less changes is as follows:

```
        public BaseCefBrowser(Func<CefRequestContext> requestContextFactory = null)
        {
            if (!CefRuntimeLoader.IsLoaded)
            {
                CefRuntimeLoader.Load();
            }
            
            ...

            RequestContext = requestContextFactory?.Invoke();
        }

File: CefGlue.Common/BaseCefBrowser.cs [hunk 20]
```diff
@@ -33,11 +33,11 @@ public BaseCefBrowser()
 
             if (CefRuntimeLoader.IsOSREnabled)
             {
-                _adapter = new CommonOffscreenBrowserAdapter(this, nameof(BaseCefBrowser), CreateOffScreenControlHost(), CreatePopupHost(), _logger);
+                _adapter = new CommonOffscreenBrowserAdapter(this, nameof(BaseCefBrowser), CreateOffScreenControlHost(), CreatePopupHost(), _logger, cefRequestContextFactory());
```
joaompneves, line 36: watch out the nulls
```suggestion
                _adapter = new CommonOffscreenBrowserAdapter(this, nameof(BaseCefBrowser), CreateOffScreenControlHost(), CreatePopupHost(), _logger, cefRequestContextFactory);
```

CefGlue.Common/BaseCefBrowser.cs
```diff
@@ -33,11 +33,11 @@ public BaseCefBrowser()
...
             } 
             else
             {
-                _adapter = new CommonBrowserAdapter(this, nameof(BaseCefBrowser), CreateControl(), _logger);
+                _adapter = new CommonBrowserAdapter(this, nameof(BaseCefBrowser), CreateControl(), _logger, cefRequestContextFactory());
```
joaompneves, line 40:
```suggestion
                _adapter = new CommonBrowserAdapter(this, nameof(BaseCefBrowser), CreateControl(), _logger, cefRequestContextFactory);
```

CefGlue.Common/CommonBrowserAdapter.cs
```diff
@@ -33,13 +33,14 @@ internal class CommonBrowserAdapter : ICefBrowserHost, IDisposable
 
         private object _disposeLock = new object();
 
-        public CommonBrowserAdapter(object eventsEmitter, string name, IControl control, ILogger logger)
+        public CommonBrowserAdapter(object eventsEmitter, string name, IControl control, ILogger logger, CefRequestContext cefRequestContext = null)
```
joaompneves, line 36:
```suggestion
        public CommonBrowserAdapter(object eventsEmitter, string name, IControl control, ILogger logger, Func<CefRequestContext> cefRequestContextFactory = null)
```

CefGlue.Common/CommonBrowserAdapter.cs
```diff
@@ -33,13 +33,14 @@ internal class CommonBrowserAdapter : ICefBrowserHost, IDisposable
...
         {
             _eventsEmitter = eventsEmitter;
             _name = name;
             _logger = logger;
 
             Control = control;
+            RequestContext = cefRequestContext;
```
joaompneves, line 43:
```suggestion
            RequestContext = cefRequestContextFactory?.Invoke();
```

CefGlue.Common/BaseCefBrowser.cs
Wiks00: Sorry, my bad, was too hasty 😅

CefGlue.Avalonia/AvaloniaCefBrowser.cs
```diff
@@ -17,6 +18,10 @@ static AvaloniaCefBrowser()
             }
         }
 
+        public AvaloniaCefBrowser(Func<CefRequestContext> cefRequestContextFactory = null)
```
joaompneves, line 21: Dont forget to also change the wpf version... it might not be building

Wiks00: The build runs without problems, however, due to my interest in Avalonia, I don't quite know how the WPF library works

CefGlue.Common/BaseCefBrowser.cs
dmitry-azaraev: This is just points to think about, not suggestions to code, nor suggestions to PR:

@joaompneves 
Does Avalonia assumes what Main thread is also always UI thread?
Does CefRuntimeLoader enforces what CEF get initialized on right (Main) thread.

I'm strongly believe what CEF can be initialized on any thread actually, but also this will be mistake - actually it must be initialized only on process's main thread. Everything else is most likely will behave badly and undebuggable in some cases. (On another side - I guess, other platforms requires Main == UI thread anyway, so I guess what this invariant is always true, even on Windows.) Generally CEF/Chromium code should be free from this, but code always have some hidden assumptions, and even it feels like work, did not mean what it works properly. This is why it is easier to init CEF on app startup, just because it is how chrome designed at all (so less cumulative risks).

For one of my projects (windows only and based on self-found mini UI framework (only simple window management and tray menu)), I'm implemented lazy CEF loading only when control is really shown. I'm was focused to use CEF's native message loop, so I'm implemented "replaceable" message loop. So, generally it was close to code above, but triggering initialization happened from paint request, rather than just class creation and custom message loops of course.

The problem what on cold host/os - loading 100+MiB binary is huge traffic even for SSD and lot of work for loader, and might took 100-500-1000ms+ (depending on hardware and OS-related factors). I'm probably even perform dll load on one thread, but initialize on main thread (however, this was years ago, so i'm not sure). Anyway, i'm tried to achieve "true" lazy loading while minimizing as much as possible application window unresponsiveness. Don't underestimate loading times: there is lot of situations when (especially cheap) SSDs go into GC on reads (and oldy HDDs never do it, and behave much better on any load). Your host just perform some writes, and now you just read dll... and, you may get whopping 5-10 seconds lock. Chromium itself uses DLL preloading on Windows very early in `chrome.exe` shim, but for CEF this never true (because client code plays this role of shim). Just to point, what true laziness is trickier than it should, and not so naive as in code above...

Also, if Avalonia allows to create controls on thread another than UI - then `Func<CefRequestContext>` is not enough, because it can be created only at CEF UI thread. Also it might require some setup before use, which currently achievable with sync calls, but this generally not necessary to be true always. E.g. `Func<ValueTask<CefRequestContext>>` can be not bad alternative, allowing performing some async work inside, if it needed.

CefGlue.Avalonia/AvaloniaCefBrowser.cs
joaompneves: just make the same change here: https://github.com/OutSystems/CefGlue/blob/main/CefGlue.WPF/WpfCefBrowser.cs
and then build the project
both avalonia and wpf implementations work almost the same, since they share more than 98% of the code

CefGlue.Common/BaseCefBrowser.cs
joaompneves: >>> Does Avalonia assumes what Main thread is also always UI thread? 
Avalonia assumes the thread where it is initialized to be the UI Thread, so not necessarily the main thread. But likewise WPF, all controls must be created on the UI thread, otherwise and exception is raised.

>>> Does CefRuntimeLoader enforces what CEF get initialized on right (Main) thread.
It doesn't. Probably it should, but thats specific to each UI Framework (Avalonia/WPF), and that piece of code is UI agnostic.
Since CefRuntimeLoader.Load(), calls the cef initialization and according to the [docs](https://github.com/OutSystems/CefGlue/blob/db4b1e74c891f121abc1086126bf217580ecbae7/CefGlue/CefRuntime.cs#L206) it must be called on the main thread. Not doing so, leads to very strange things, because one of the things that cef does is to hook up with the process message loop. On Mac its even more problematic, when cef and the UI thread dont coincide.

In the general case of Desktop apps, and to speedup things, we initialize CEF as soon as we believe it will be used, ie, on the first control instantiation. That way we can save a few milieseconds on the first load, if we considering that not all the work will be done on the UI thread.

@dmitry-azaraev I'm glad you are engaging with this PR, but would be great that you payed some more attention 3 years ago when I first opened the PR on the CefGlue gitlab repo. We did alot of work to create an Avalonia port that can run on Mac and Windows, we separated the browser process with the hosting process, we fixed alot of memory leaks that the original CefGlue implementation has, we introduced a powerfull (strongly typed) ways to comunicate with the browser (js evaluation for .net side, and calling .net code form the browser), we created lots of unit tests... Moreover we have several thousands of users using this every day on an enterprise grade product, and at this point I'm confident with the quality of what we have.
I understand that your interests rely mostly on the core, and alot of our work was done on top of that, but IMO it would be better for the community if we joined efforts, you taking care of the Core (most of our changes on the Core were related with memory leaks fixes) and we could take care of the Avalonia/WPF bindings. My 2 cents.

dmitry-azaraev: @joaompneves I'm closed all oldy PRs. I'm can't understand them. There is too much code in once. I'm gladly accept any sensible improvements (or some things which you might require, but it is clear what there is not too many things), but I'm can't accept to merge just bunch of unrelated code. If you maintain updates and need regularly update some things in core - let's discuss them and push them. It should not be a problem. Let's do it if it helps you. Just split core from uncore code, because uncore code is yours anyway and can live separately without harrasing each-other.

As for very "big" plan, what might help community - it would be nice to create new bindings. What's because any CefGlue alternative actually have lot of issues. I'm saying not because they are competitor projects, no. I'm saying because I'm having some unexpected journey with almost all of them as well, and all of them very far from ideal. One of main problems of CefGlue is maintenance hell. And they are solveable. I'm full of new ideas for modern days, it was already started 2-3 times in past years, but never finished (and should be started from zero again, because I'm any way did not remember anything).

Also I'm living in warzone and last year was especially hard, and continues to be hard. Previous years actually was not been too easier, but at least more predictable. Also, just now, I'm not really active CEF user, and by so, I'm doing other Chromium related stuff, and by so, this did not push me actively to change something. CefGlue is very-very-very old project (born in CEF1 era). Few days ago I'm touch generation script because otherwise 108 is was not possible, and this is challenge, because it is dumb script, and everything done is somehow badly (many generators are tricky, but it is tricky twice). Huh. Nowaday, I'm never will write code generator in python.

And main problem with new bindings, what I'm pick (you can read as hire) people(s) for it, and they leave from project without progress. There is was always lot of reasons to do it from 2014, but still this yielded to no progress.

So, just now I'm even not sure what I'm really want to share this work/responsibility with someone, but I'm understand what I'm personally not so interested to make it in only mine hands, as it was in past years. On another side, because I'm know how it must be done in better way - makes me feel bad, because I'm want make it nice.

So, just now, I'm just support it in term of update it time-to-time, not even strictly with every version. Just once in one or two months.

I'm hope what eventually I'm back to more positive mood, and can bring more. But just now this almost impossible. Sorry.

CefGlue.WPF/WpfCefBrowser.cs
```diff
@@ -12,6 +13,13 @@ namespace Xilium.CefGlue.WPF
     public class WpfCefBrowser : BaseCefBrowser
     {
         public WpfCefBrowser()
+            : base(null)
+        {
+            KeyboardNavigation.SetAcceptsReturn(this, true);
+        }
```
joaompneves, line 19:
```suggestion
        public WpfCefBrowser() : this(null) { }
```

Wiks00:
> Thank you for your contribution. This will be available in nuget version 106.5249.4

Thanks for the help.

PR #76, opened 2023-02-24
RICT-5081 :: Adding localized packages folder

osesantos:


joaompneves (changes requested):
whats going on with this PR ... 55 files changed!

PR #78, opened 2023-02-28
Fixing BrowserProcess packaging

osesantos:
`PublishApp` was called after the `CopyProjectReferencesToPackage`, resulting in a NuGet package without the BrowserProcess files in it.


CefGlue.Avalonia/CefGlue.Avalonia.csproj
```diff
@@ -9,6 +9,7 @@
     <PackageId>CefGlue.Avalonia</PackageId>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
+    <PlatformTarget>AnyCPU</PlatformTarget>
```
joaompneves, line 12: is this change required... this doesn't look good because cefglue is x64 only... should not be built as any cpu

Nuget.config
```diff
@@ -1,5 +1,16 @@
...
+	<config>
+        <add key="repositoryPath" value="packages" />
+        <add key="globalPackagesFolder" value="packages" />
+    </config>
+	<packageRestore>
+		<!-- Opt-In both Package Restore, and MSBuild-Integrated Package Restore, does not make MSBuild restore automatic -->
+		<add key="enabled" value="True" />
+		<!-- Opt-In Automatic Package Restore in Visual Studio -->
+		<add key="automatic" value="True" />
+	</packageRestore>
```
joaompneves, line 12: is this change related/needed?

fix identation

Directory.Packages.props
```diff
@@ -16,9 +16,9 @@
     </ItemDefinitionGroup>
 
     <ItemGroup>
-        <PackageVersion Include="Avalonia" Version="0.10.17" />
-        <PackageVersion Include="Avalonia.Desktop" Version="0.10.17" />
-        <PackageVersion Include="Avalonia.Diagnostics" Version="0.10.17" />
+        <PackageVersion Include="Avalonia" Version="0.10.15" />
+        <PackageVersion Include="Avalonia.Desktop" Version="0.10.15" />
+        <PackageVersion Include="Avalonia.Diagnostics" Version="0.10.15" />
```
joaompneves, line 21: please revert

Directory.Build.targets
```diff
@@ -5,4 +5,12 @@
     <DocumentationFile />
   </PropertyGroup>
 
+  <Target Name="PublishBrowserProcessApp" Inputs="(ProjectDir)../.publishbpapp.stamp" Outputs="(ProjectDir)../.publishbpapp.stamp">
```
joaompneves, line 8: this target should be place where its needed

joaompneves:
```suggestion
  <Target Name="PublishBrowserProcessApp" Inputs="(ProjectDir)..\.publishbpapp.stamp" Outputs="(ProjectDir)..\.publishbpapp.stamp">
```

Directory.Build.targets
```diff
@@ -5,4 +5,12 @@
...
+    <!-- Publish both platforms on Windows for packaging -->
+    <Message Text="Publishing BrowserProcess... [$(ProjectPath)]" Importance="High" />
+    <MSBuild Projects="$(ProjectDir)../CefGlue.BrowserProcess/CefGlue.BrowserProcess.csproj" Targets="Publish" Properties="Configuration=$(Configuration);TargetFramework=$(TargetFramework);PublishTrimmed=True;RuntimeIdentifier=win-x64" />
```
joaompneves, line 11: use \ ... / is used when using a prop which msbuild doesn't recognize as path

Directory.Build.targets
```diff
@@ -5,4 +5,12 @@
...
+    <MSBuild Projects="$(ProjectDir)../CefGlue.BrowserProcess/CefGlue.BrowserProcess.csproj" Targets="Publish" Properties="Configuration=$(Configuration);TargetFramework=$(TargetFramework);PublishTrimmed=True;RuntimeIdentifier=osx-x64" />
+    <Touch Files="$(ProjectDir)../.publishbpapp.stamp" AlwaysCreate="true"/>
```
joaompneves, line 13: this stamp file should go to the obj folder... its the appropriate place for cache files

joaompneves (changes requested):
I would like to understand why this stopped working. The upgrade shouldn't be related, only possibly the move from .net 5 to .net 6

Nuget.config
osesantos: using the local packages folder is not very good for project isolation and that was causing us a hard time to troubleshoot problems and test changes locally

Directory.Packages.props
osesantos: no, we agreed that the avalonia version would be upgraded on his own pr and version

Directory.Build.targets
osesantos: I disagree because this target acts on the BrowserProcess project but is being called by the Common project which means that from an organization's point of view, it should be placed in a place that is common to both projects.

CefGlue.Avalonia/CefGlue.Avalonia.csproj
joaompneves: thats an SS problem... not cefglue

Directory.Build.targets
osesantos: actually, I think it does, .net6 changed that, and the proof of that is the tests I did both on win, mac, and the pipeline. 
But to keep consistency with other targets on this solution I'll change it

Directory.Packages.props
joaompneves: but its a downgrade

Directory.Build.targets
joaompneves: that was not a problem before... it was placed on the browser process project... still not clear for me the need of these so many changes

Directory.Build.targets
joaompneves: do you have a documented proof? that would be a major breaking change which I0m a bit skeptical that they would introduce that

Directory.Build.targets
osesantos: ok even so I disagree because this target is common to all projects, creating the stamp on the build folder of the BrowserProcess feels to me like a hack. However, I don't foresee any impact on the final output of the nuget so Im ok on moving that it to the build

Directory.Build.targets
osesantos: https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats
image
By using / we are both enabling building on win and mac

Directory.Build.targets
osesantos: yes, it was on the browser process, but the target that is on the common is dependent on this one to actually copy this target output into the nuget. This way the target is available on the Common project

Directory.Packages.props
osesantos: sure, we were seeing some build processes being left running even after the build was finished alongside with Avalonia dlls being locked. I want to avoid more external changes and validate these ones first. btw, the upgrade was made by @cangosta on his pr

CefGlue.Avalonia/CefGlue.Avalonia.csproj
osesantos: its also a cefglue problem because we dont want SS platform to be locked to the one used by cefglue.

Directory.Build.targets
osesantos: image
sorry but I can't do it... I guess the build folder is not yet available when this target executes

CefGlue.Avalonia/CefGlue.Avalonia.csproj
joaompneves: It needs to be locked.. it simply wont work. Cefglue doesnt support x86 architecture anymore. Running on x86 will crash right away.

osesantos: fair point, however, we will have to revisit this in the future since we will want to native ARM builds

Directory.Packages.props
osesantos: I locally tested it and reverting the version allowed me to get it working
also, by running the build via the terminal we can see on the task manager that some processes are left running, if we try to git clean then you can see that Avalonia.task.dll is locked. Ending those processes unlocks the dll

Directory.Build.targets
osesantos: sorry, I think I'm missing something... if I can use / or \ for paths and should use / for text to allow cross-platform compilation, why not just use /  everywhere?

joaompneves: because editors (VS, Rider) prefer \ ... that would be a lost game trying to fight with that... so the "best" rule is what I described

Directory.Packages.props
osesantos: reverted back to 17

Nuget.config
osesantos: identation fixed

Directory.Build.targets
```diff
@@ -5,4 +5,12 @@
     <DocumentationFile />
   </PropertyGroup>
 
+  <Target Name="PublishBrowserProcessApp" Inputs="$(ProjectDir)..\CefGlue.BrowserProcess\obj\.publishbpapp.stamp" Outputs="$(ProjectDir)..\CefGlue.BrowserProcess\obj\.publishbpapp.stamp">
+    <!-- Publish both platforms on Windows for packaging -->
```
alvesmiguel1, line 9: We are now publishing in both operating systems, right?

alvesmiguel1: Since you removed the `IsWindows` property

.gitignore
```diff
@@ -38,3 +38,6 @@ NuGet/output/
 packages/
 CefGlue.Interop.Gen/Classes.Handlers.tmpl/
 CefGlue.Interop.Gen/Classes.Proxies.tmpl/
+
+#Stamps
+**/*.stamp
```
joaompneves, line 43: not needed since stamp files are inside obj

Directory.Build.targets
```diff
@@ -5,4 +5,12 @@
...
+    <Message Text="Publishing BrowserProcess... [$(ProjectPath)]" Importance="High" />
+    <MSBuild Projects="$(ProjectDir)..\CefGlue.BrowserProcess\CefGlue.BrowserProcess.csproj" Targets="Publish" Properties="Configuration=$(Configuration);TargetFramework=$(TargetFramework);PublishTrimmed=True;RuntimeIdentifier=win-x64" />
+    <MSBuild Projects="$(ProjectDir)..\CefGlue.BrowserProcess\CefGlue.BrowserProcess.csproj" Targets="Publish" Properties="Configuration=$(Configuration);TargetFramework=$(TargetFramework);PublishTrimmed=True;RuntimeIdentifier=osx-x64" />
+    <Touch Files="$(ProjectDir)..\CefGlue.BrowserProcess\obj\.publishbpapp.stamp" AlwaysCreate="true"/>
+  </Target>
```
joaompneves, line 14:
```suggestion
  <PropertyGroup>
      <BrowserProcessAppPublishStamp>$(IntermediateOutputPath)BrowserProcessAppPublish.stamp</BrowserProcessAppPublishStamp>
  </PropertyGroup>
  
  <Target Name="PublishBrowserProcessApp" Inputs="$(BrowserProcessAppPublishStamp)" Outputs="$(BrowserProcessAppPublishStamp)">
    <!-- Publish both platforms on Windows for packaging -->
    <Message Text="Publishing BrowserProcess... [$(ProjectPath)]" Importance="High" />
    <MSBuild Projects="$(ProjectDir)..\CefGlue.BrowserProcess\CefGlue.BrowserProcess.csproj" Targets="Publish" Properties="Configuration=$(Configuration);TargetFramework=$(TargetFramework);PublishTrimmed=True;RuntimeIdentifier=win-x64" />
    <MSBuild Projects="$(ProjectDir)..\CefGlue.BrowserProcess\CefGlue.BrowserProcess.csproj" Targets="Publish" Properties="Configuration=$(Configuration);TargetFramework=$(TargetFramework);PublishTrimmed=True;RuntimeIdentifier=osx-x64" />
    <Touch Files="$(BrowserProcessAppPublishStamp)" AlwaysCreate="true"/>
  </Target>
```

osesantos: why this change to IntermediateOutputPath? won't that path depend on the project that is calling it? that was not what I wanted, the stamp is to avoid the PublishBrowserProcessApp more than once when called by multiple projects

joaompneves: the target will be called by the common project, hence it will store the stamp on the common project intermediate folder. Thats why I also suggested the target to be moved to the common project... its where it makes sense.

joaompneves: Moreover have you tried building the whole project, then making a change on the browser process project and build the demo again? will the browser process be built again and take the new change?

Directory.Build.targets
joaompneves: previously we only built both when on windows, because the azure pipeline runs on windows... any good reason for that change?

Directory.Build.targets
osesantos: > Moreover have you tried building the whole project, then making a change on the browser process project and building the demo again? will the browser process be built again and take the new change?
the objective of this target is not to be called each time the BrowserProcess is built, but instead be called when we are creating the Common NuGet, this is because we are shipping the BrowserProcess inside the Common instead of shipping it on its own NuGet.
To answer your question, yes the BrowserProcess will be built again, will this target run? no, we have to delete the stamp or create a clean target to do that. That is why I still think that having it on the root was better to identify it and remove it if needed.

Directory.Build.targets
osesantos: if we want to change the pipeline to ubuntu, for instance, this will be a blocker. I don't see any advantage in having that type of condition if we always want the NuGet to have both win and osx available regardless of the OS that is building it.

Directory.Build.targets
joaompneves: thats big deal... because if you make changes you wont be able to see those changes. that needs to be fixed

Directory.Build.targets
joaompneves: is that a problem? is that what you're trying to fix? stick with the minimal changes required, please. unless you're willing to buy new problems

CefGlue.Common/CefGlue.Common.csproj
```diff
@@ -39,6 +39,11 @@
             <IncludeAssets>CefGlue.dll</IncludeAssets>
         </ProjectReference>
 
+        <ProjectReference Include="..\CefGlue.BrowserProcess\CefGlue.BrowserProcess.csproj">
```
joaompneves, line 42:
```suggestion
        <!-- Forces BrowserProcess build before CopyProjectReferencesToPackage target run -->
        <ProjectReference Include="..\CefGlue.BrowserProcess\CefGlue.BrowserProcess.csproj">
```

joaompneves (approved):

PR #100, opened 2023-07-18
Support CefGlue nuget consumers publish

alvesmiguel1:


CefGlue.Common/build/CefGlue.Common.props
```diff
@@ -3,16 +3,23 @@
...
+        <CefGlueTargetPlatform Condition="'$(Platform)' == 'x64'">win-x64</CefGlueTargetPlatform>
+        <CefGlueTargetPlatform Condition="'$(Platform)' == 'ARM64'">win-arm64</CefGlueTargetPlatform>
+        <CefGlueTargetPlatform Condition="'$(CefGlueTargetPlatform)' == ''">win-x64</CefGlueTargetPlatform> <!-- fallback -->
     </PropertyGroup>
-    <PropertyGroup Condition="$([MSBuild]::IsOSPlatform('OSX'))">
-        <CefGlueTargetPlatform Condition="'$(Platform)' == 'x64' or '$(RuntimeIdentifier)' == 'osx-x64'">osx-x64</CefGlueTargetPlatform>
-        <CefGlueTargetPlatform Condition="'$(Platform)' == 'ARM64' or '$(RuntimeIdentifier)' == 'osx-arm64'">osx-arm64</CefGlueTargetPlatform>
-        <CefGlueTargetPlatform Condition="'$(CefGlueTargetPlatform)' == '' and '$(RuntimeIdentifier)' == ''">osx-x64</CefGlueTargetPlatform> <!-- fallback -->
+    
+    <PropertyGroup Condition="'$(RuntimeIdentifier)' == '' and $([MSBuild]::IsOSPlatform('OSX'))">
```
igor-ruivo, line 13: this is just for local builds, right?
because in the pipeline, during the build step (where we also don't have runtimeidentifier filled), this is always going to be false because we use a windows agent, right?

igor-ruivo (approved):

CefGlue.Common/build/CefGlue.Common.props
alvesmiguel1: >this is just for local builds, right?

Yap

>because in the pipeline, during the build step (where we also don't have runtimeidentifier filled), this is always going to be false because we use a windows agent, right?

Yes, publish would fail

PR #109, opened 2023-09-20
RDOIBT-1838 - CefGlue: Duplicate key reference

OS-filipecampos:


mmv (approved):

alvesmiguel1 (approved):

joaompneves:
@mmv shouldn't we follow-up on this... what was your intent in the first place? won't this revert it?

mmv:
The original intent was to gain performance while mimicking the reference counting behavior that Cef does on the native side, where it uses atomic increment/decrement operations and frees the memory once the reference counter reaches zero. Atomic counters are significantly more performant than mutex locks, and that would get relevant in code that gets called a lot of times. The native side will call `add_ref` and `release` every time a pointer is created for the value and every time the pointer gets dropped. If the counter drops to zero, that means there's zero references on the native side, and so it should be safe to let the GC collect the thing on our side.

The issue that was overlooked was that this only works under the assumption that a handler would be created once and Cef would tell us when it's done, but the managed side can deliver the same managed handler to multiple Cef things, and we may have a case where one Cef thing is about to release the handler while another is about to start using it. In those cases we can have two threads in the managed side doing the counter check but then following on and taking inconsistent actions.

So in the end between performance and stability we should chose stability. We can revisit this later to add slightly more complex logic (eg: use atomic counters and only lock when we're about to release the object) but that would need some care to ensure that no race conditions like this one would still be possible.

joaompneves (commented):
missing raise version

alvesmiguel1 (approved):

PR #126, opened 2023-12-19
120.1.8+ge6b45b0+chromium-120.0.6099.109

JMano:


Directory.Build.props
```diff
@@ -3,15 +3,15 @@
     <PropertyGroup>
         <DotnetVersion>net6.0</DotnetVersion>
         <LangVersion>latest</LangVersion>
-        <CefRedistVersion>117.2.4</CefRedistVersion>
-        <CefRedistOSXVersion>117.2.4</CefRedistOSXVersion>
+        <CefRedistVersion>120.1.8</CefRedistVersion>
+        <CefRedistOSXVersion>120.1.8</CefRedistOSXVersion>
```
alvesmiguel1, line 7: 💪

Nuget.config
```diff
@@ -15,4 +15,10 @@
     <packageSources>
         <add key="Nuget" value="https://api.nuget.org/v3/index.json" />
     </packageSources>
+
+    <packageSourceMapping>
```
OS-conradogoncalves, line 19: Was this change intended?

OS-conradogoncalves (approved):
Only one question regarding a change in Nuget.config. Otherwise LGTM 👍

osesantos (approved):

Nuget.config
JMano: Not really but it doesn't hurt to be here 😅

CefGlue.Common/InternalHandlers/CommonCefLifeSpanHandler.cs
```diff
@@ -40,5 +40,17 @@ protected override bool OnBeforePopup(CefBrowser browser, CefFrame frame, string
...
+
+        protected override void OnBeforeDevToolsPopup(CefBrowser browser, CefWindowInfo windowInfo, ref CefClient client, CefBrowserSettings settings, ref CefDictionaryValue extraInfo, ref bool useDefaultWindow)
+        {
+            if (_owner.LifeSpanHandler != null)
+            {
+                _owner.LifeSpanHandler?.HandleBeforeDevToolsPopup(browser, windowInfo, ref client, settings, ref extraInfo, ref useDefaultWindow);
+            }
+            else
+            {
+                base.OnBeforeDevToolsPopup(browser, windowInfo, ref client, settings, ref extraInfo, ref useDefaultWindow);
```
joaompneves, line 52: why isnt the base being called always?

CefGlue.Common/CommonBrowserAdapter.cs
```diff
@@ -314,6 +318,31 @@ public bool CreateBrowser(int width, int height)
...
+
+        public bool CanExecuteChromeCommand(int commandId)
+        {
+            return BrowserHost?.CanExecuteChromeCommand(commandId) ?? false;
+        }
+
+        public void ExecuteChromeCommand(int commandId, CefWindowOpenDisposition windowOpenDisposition)
+        {
+            BrowserHost?.ExecuteChromeCommand(commandId, windowOpenDisposition);
+        }
```
joaompneves, line 344: is there a reason for these being introduced?

JMano: No reason aside from trying to keep this in sync with the additions to `cef_browser.h`, similar to other methods that were already here.

You weren't available when this was done so I couldn't confirm with anyone else what should or should not be added 😅 If you feel this is unnecessary or should be removed, I'm more than glad to open a new PR reverting this.

CefGlue.Common/InternalHandlers/CommonCefLifeSpanHandler.cs
JMano: I think this was a mistake on my side. I'll create a PR to ensure `base.OnBeforeDevToolsPopup` is always called.

CefGlue.Common/CommonBrowserAdapter.cs
joaompneves: we only introduce if necessary, there lots of other things not available, and we should only introduce things on a need basis, otherwise its more effort to maintain

joaompneves: these specific things I dont see to point to be available

PR #136, opened 2024-05-13
Support .NET 8 and multitargeting

alvesmiguel1:
Support both .NET 6 and .NET 8

**Pipeline:**

image

**Nugets:**

image

### DIFF
--- a/CefGlue.Demo.WPF/CefGlue.Demo.WPF.csproj
+++ b/CefGlue.Demo.WPF/CefGlue.Demo.WPF.csproj
@@ -2,7 +2,7 @@
 
     <PropertyGroup>
         <OutputType>WinExe</OutputType>
-        <TargetFrameworks>net6.0-windows;net8.0-windows</TargetFrameworks>
+        <TargetFramework>$(DefaultTargetDotnetVersion)-windows</TargetFramework>
         <AssemblyName>Xilium.CefGlue.Demo.WPF</AssemblyName>
         <RootNamespace>Xilium.CefGlue.Demo.WPF</RootNamespace>
         <UseWPF>true</UseWPF>
@@ -34,7 +34,7 @@
 
 
     <ItemGroup>
-        <ProjectReference Include="..\CefGlue.WPF\CefGlue.WPF.csproj" />
+        <ProjectReference Include="..\CefGlue.WPF\CefGlue.WPF.csproj" SetTargetFramework="TargetFramework=$(TargetFramework)" />
         <ProjectReference Include="..\CefGlue\CefGlue.csproj" />
         <ProjectReference Include="..\CefGlue.Common\CefGlue.Common.csproj" />
         <ProjectReference Include="..\CefGlue.Common.Shared\CefGlue.Common.Shared.csproj" />
--- a/CefGlue.Demo.WPF/CefGlue.Demo.WPF.csproj
+++ b/CefGlue.Demo.WPF/CefGlue.Demo.WPF.csproj
@@ -2,7 +2,7 @@
 
     <PropertyGroup>
         <OutputType>WinExe</OutputType>
-        <TargetFramework>$(DotnetVersion)-windows</TargetFramework>
+        <TargetFrameworks>$(TargetDotnetVersionsWindows)</TargetFrameworks>
         <AssemblyName>Xilium.CefGlue.Demo.WPF</AssemblyName>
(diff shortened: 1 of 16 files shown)


Directory.Build.props
```diff
@@ -1,7 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project>
     <PropertyGroup>
-        <DotnetVersion>net6.0</DotnetVersion>
+        <TargetDotnetVersions>net6.0;net8.0</TargetDotnetVersions>
```
joaompneves, line 4: can we just stick with one version?

alvesmiguel1: That's the goal, this will be temporary until we are officially releasing a .NET 8 version. We would be just temporarily multi-targeting two .NET versions, I believe there are no major changes required in the repo to achieve that, neither in consumer side. And this would avoid having one more branch (we still have the Avalonia one)

tbragaf (approved):

Vivokas20 (dismissed):

Vivokas20 (dismissed):

Directory.Build.props
```diff
@@ -1,7 +1,8 @@
...
+        <TargetDotnetVersionsWindows>net6.0-windows;net8.0-windows</TargetDotnetVersionsWindows>
```
joaompneves, line 5: split... platform and versions

alvesmiguel1: I tried to do that by having something like this in WPF projects:

```
<NetVersionSuffix>-windows</NetVersionSuffix>
<TargetFrameworks>
        $(TargetDotnetVersions.Replace(';', $(NetVersionSuffix) + ';'))$(NetVersionSuffix)
</TargetFrameworks>
```

But I'm having this error:

image

alvesmiguel1: I think that MSBuild does not behave well with TargetFrameworks conditional to some other properties. The values are the same

alvesmiguel1: Unfortunately I think we cannot achieve that

alvesmiguel1: I was able to overcome only by defining a default target framework for CefGlue.Demo.WPF project

Directory.Build.props
```diff
@@ -1,7 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project>
     <PropertyGroup>
-        <DotnetVersion>net6.0</DotnetVersion>
+        <DefaultTargetDotnetVersion>net8.0</DefaultTargetDotnetVersion>
+        <TargetDotnetVersions>net6.0;net8.0</TargetDotnetVersions>
```
joaompneves, line 5:
```suggestion
        <TargetDotnetVersions>net6.0;$(DefaultTargetDotnetVersion)</TargetDotnetVersions>
```

CefGlue.Demo.WPF/CefGlue.Demo.WPF.csproj
```diff
@@ -2,12 +2,15 @@
 
     <PropertyGroup>
         <OutputType>WinExe</OutputType>
-        <TargetFramework>$(DotnetVersion)-windows</TargetFramework>
+        <DotnetVersionSuffix>-windows</DotnetVersionSuffix>
+        <TargetFramework>$(DefaultTargetDotnetVersion)$(DotnetVersionSuffix)</TargetFramework>
+        <TargetFrameworks>$(TargetDotnetVersions.Replace(";", "$(DotnetVersionSuffix);"))$(DotnetVersionSuffix)</TargetFrameworks>
```
tbragaf, line 7: what happens if the string ends in ";"? this would fail, but not sure if it's allowed in dotnet

alvesmiguel1: It actually didn't work, so I reverted it

Directory.Build.props
alvesmiguel1: I needed to revert the change because it didn't work 😐

Directory.Build.props
alvesmiguel1: Actually, it didn't work... I reverted it

joaompneves: please simplify... we dont need to support wpf .net v6

alvesmiguel1: okay, I'll add a default target

alvesmiguel1: Done

joaompneves (approved):

PR #137, opened 2024-05-15
Publish .NET 8 BrowserProcess as a self-contained app

alvesmiguel1:

PR #149, opened 2024-06-28
RDEV-6642 Make CefGlue SelfContained

Vivokas20:
While testing ODC running on dotnet 8 we noticed that CefGlue needed the dotnet8 sdk installed to be able to run.
This was due to a breaking change in dotnet 8 which made some apps no longer self-contained by default: https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/8.0/runtimespecific-app-default

### DIFF
--- a/CefGlue.BrowserProcess/CefGlue.BrowserProcess.csproj
+++ b/CefGlue.BrowserProcess/CefGlue.BrowserProcess.csproj
@@ -8,7 +8,7 @@
     <RuntimeIdentifiers>osx-x64;win-x64;osx-arm64;win-arm64</RuntimeIdentifiers>
     <RunPostBuildEvent>OnOutputUpdated</RunPostBuildEvent>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
-    <PublishCommonConfig>Configuration=$(Configuration);Platform=$(Platform);TargetFramework=$(TargetFramework);IsPublishing=True;PublishTrimmed=True;RuntimeIdentifier=</PublishCommonConfig>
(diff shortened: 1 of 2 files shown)


joaompneves (commented):
https://github.com/OutSystems/CefGlue/pull/137#discussion_r1601499137

Vivokas20:
@joaompneves Just for some more context this was the error we got when trying to run Studio on net8
image

After analyzing your comment me and @alvesmiguel1 went and check the Publish folder generated while building the CefGlue repo in ReleaseWPFAvalonia mode.

We noticed that the folder generated with **net6** and **without** `self-contained` had roughly 80,5 MB and the one with **net8** **with** `self-contained=True` had 82,2 MB. We also noticed that the libcef was not present on the Publish folder with self-contained=True. For this reason we believe it's ok to follow this implementation. Is there anything we might be missing?

Just for comparison building using net8 without `self-contained=True` yield a folder of only 2,1 MB which is clearly missing files we need.

joaompneves:
> @joaompneves Just for some more context this was the error we got when trying to run Studio on net8 image
> 
> After analyzing your comment me and @alvesmiguel1 went and check the Publish folder generated while building the CefGlue repo in ReleaseWPFAvalonia mode.
> 
> We noticed that the folder generated with **net6** and **without** `self-contained` had roughly 80,5 MB and the one with **net8** **with** `self-contained=True` had 82,2 MB. We also noticed that the libcef was not present on the Publish folder with self-contained=True. For this reason we believe it's ok to follow this implementation. Is there anything we might be missing?
> 
> Just for comparison building using net8 without `self-contained=True` yield a folder of only 2,1 MB which is clearly missing files we need.

Can compare with the previous version before introducing the .net 8 changes (can be [this](https://github.com/OutSystems/CefGlue/commit/3a03c28516d4578c7bf4b461c429483e20fd4d9d) commit)?

Vivokas20:
> this

Just checkout that commit, deleted bin and built again, the resulting folder also had the expected 80,5 MB
image

joaompneves (approved):

PR #159, opened 2024-09-02
Single targeting .NET 8 in CefGlue

alvesmiguel1:
Also reverted some changes to support multitarget, as it is not functioning properly

### DIFF
--- a/CefGlue.CopyLocal.props
+++ b/CefGlue.CopyLocal.props
@@ -1,31 +1,30 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project>
     <Import Project="CefGlue.Packages.props" />
+
+    <PropertyGroup>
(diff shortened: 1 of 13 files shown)


CefGlue.CopyLocal.props
```diff
@@ -1,31 +1,30 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project>
     <Import Project="CefGlue.Packages.props" />
+
+    <PropertyGroup>
```
joaompneves, line 5: whats this for?

alvesmiguel1: Wrongly re-added this again (it was like this before multi-target support). Didn't notice it was already defined below

joaompneves (dismissed):

joaompneves (approved):

PR #160, opened 2024-09-02
Dump test details

joaompneves:


alvesmiguel1 (dismissed):

alvesmiguel1:
The failing test is passing locally in my VM:

image

joaompneves:
> The failing test is passing locally in my VM:
> 
> image

yeah but the test is flacky, we need to understand why

PR #170, opened 2024-10-15
Binary inter process communication (IPC)

piotrek1668:
@joaompneves I have created a fork and the branch with our changes. We would be very pleased if you would give us your opinion on the change. Additionally it would be nice to be able to contribute the branch as an extension to the CefGlue repository.

A good starting point is the PerformanceTests.cs in CefGlue.Tests

The test shows the difference in performance between a json and a binary serializer.

Which serializer should be used can be configured in the RegisterJavascriptObject method. By default, the serializer is still set to Json. 

This is the signature of the method:
```c#
public void RegisterJavascriptObject(object targetObject, string name, MethodCallHandler methodHandler = null, Messaging messaging = null)
{
       // Implementation
}
``` 

Here some examples how to use it:
```c#
var _dotNetObject = new BindingTestClass();

// default usecase with json serializier
Browser.RegisterJavascriptObject(_dotNetObject, "dotNetObject");

 // define json serializier explicitly
Browser.RegisterJavascriptObject(_dotNetObject, "dotNetObject", null, Messaging.Json);

 // define json serializier explicitly (without passing null value for method handler)
Browser.RegisterJavascriptObject(_dotNetObject, "dotNetObject", messaging: Messaging.Json);

// define binary serializer (MessagePack)
Browser.RegisterJavascriptObject(_dotNetObject, "dotNetObject", null, Messaging.MsgPack);

// example with a method handler
Browser.RegisterJavascriptObject(_dotNetObject, "dotNetObject", AsyncCallNativeMethod);
``` 

If you are interested what MessagePack is and how MessagePack works, look here: https://msgpack.org/

Here is the original message:

_Together with my colleagues we want to introduce an alternative binary serializer for the CefGlue library. The serializer can optionally be changed for communication between client and browser. By default, the serializer is still set to Json, but if there is a need to use a faster binary serialisiert, the user can change to use MessagePack for serialization. If large amounts of data has to be transferred, the serializer can be set to MsgPack. Depending on the data size, the transmission performance increases by a factor of 10+._

@rrattke FYI


joaompneves:
First of all thanks for contributing.

There's alot of changes here that need to be carefully reviewed. Some of the changes are unnecessary (eg: indentation) and you should try to adhere to the repo code style as much as possible to reduce the number of changes. 
I think the best approach would be to create some interfaces to be able to replace the serializer approach (on demand) and put the alternative serializer outside of this repo (as an extension). Not only that would reduce the breaking changes risk but also can be an optin, ie, people should only need to bring the Msgpack dependency if used.

piotrek1668:
Many thanks for the quick response. We will investigate the suggestion to take the extension approach and then get back to you.

PR #183, opened 2024-11-06
Disable FirstPartySets (YouTube crash)

sakya:
Due to this CEF bug https://github.com/chromiumembedded/cef/issues/3643 YouTube crash when trying to search.

### DIFF
--- a/CefGlue.Demo.Avalonia/Program.cs
+++ b/CefGlue.Demo.Avalonia/Program.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.IO;
 using Avalonia;
 using Xilium.CefGlue.Common;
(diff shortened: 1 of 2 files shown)


CefGlue.Demo.Avalonia/Program.cs
```diff
@@ -35,9 +36,13 @@ static int Main(string[] args)
                             SchemeName = "test",
                             SchemeHandlerFactory = new CustomSchemeHandler()
                         }
-                      }))
+                      },
+                      flags: [
```
joaompneves, line 40: this should be done conditionally (only when windoless is enabled) inside the CefRuntimeLoader.Initialize

joaompneves (changes requested):

CefGlue.Demo.Avalonia/Program.cs
sakya: Ok, but why only if windowless is enabled? The crash happens also with windowless disabled

joaompneves: ok, I though the problem was occurring in windowless only. in that case shouldn't be conditional.

PR #186, opened 2024-11-07
Fixed RegisterJavascriptObject

### LINKED ISSUE #185: RegisterJavascriptObject strange behaviour
I tried to use `RegisterJavascriptObject` in the browser LoadEnd event and I'm seeing a strange behaviour

```cs
        private void OnBrowserLoadEnd(object sender, LoadEndEventArgs e)
        {
            Console.WriteLine("OnBrowserLoadEnd");
            if (!browser.RegisterJavascriptObject(new BindingTestClass(), "bindingTest", AsyncCallNativeMethod))
            {
                Console.WriteLine("Failed to Register Javascript Object");
            }
        }
```

I modified `RegisterJavascriptObject` to return the boolean value from `_objectRegistry.Register'.
The strange thing is:
- When the browser loads google the object is registerer and in the console I can access the object
- If I navigate to another page (by clicking a link or writing the address) I get the error "Failed to Register Javascript Object" because the object is registered but in the console I get "Uncaught ReferenceError: bindingTest is not defined" when I try to access the object

The code that return false is this
```cs
            if (_registeredObjects.ContainsKey(name))
            {
                Console.WriteLine($"Object {name} is registered");
                return false;
            }
```
This should mean that the object is registered. Why I cannot access it?

sakya:
Fix #185 

Registered objects are only valid in the current page.
We should re-register all the objects when the page is loaded

### DIFF
--- a/CefGlue.Common/CommonBrowserAdapter.cs
+++ b/CefGlue.Common/CommonBrowserAdapter.cs
@@ -173,7 +173,7 @@ namespace Xilium.CefGlue.Common
             // Remove leading whitespace from the URL
             url = url.TrimStart();
 
-            // to play safe, load url must be called after OnBrowserCreated(CefBrowser) which runs on CefThreadId.UI, 
+            // to play safe, load url must be called after OnBrowserCreated(CefBrowser) which runs on CefThreadId.UI,
             // otherwise the navigation will be aborted
             ActionTask.Run(() =>
             {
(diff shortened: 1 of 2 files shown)


CefGlue.Common/CommonBrowserAdapter.cs
```diff
@@ -599,6 +599,8 @@ void ICefBrowserHost.HandleLoadStart(CefBrowser browser, CefFrame frame, CefTran
 
         void ICefBrowserHost.HandleLoadEnd(CefBrowser browser, CefFrame frame, int httpStatusCode)
         {
+            // Register all objects
+            _objectRegistry.SetBrowser(browser);
```
AndreBooysen, line 603: In my scenario I needed to make sure that the objects were already registered during the page load, so I called this method inside HandleLoadStart instead.

PR #188, opened 2024-11-07
Modify debug and packaging method of BrowserProcess

yangxuilyx:
BrowserProcess is released in Self-Containerd mode and requires .net runtime, so the package size will be very large (depends of win, linux, macos)
I modified debug and publish to let BrowserProcess call the runtime of the main program,modified the package structure and placed it in runtimes\native

cef runtimes is copied to the debug directory in the demo project, but the chromiumembeddedframework.runtime.win-x64 package is copied to runtimes\native by default
so modified the copy method of cef.redist to handle it in a unified way

fixed the handling of CEF locales file in demo/package on Windows platform

I passed the debug and publish tests on win-x64 and linux-x64, but I don’t have macos so there is no test.

my test project:
[https://github.com/yangxuilyx/Desktop.Ava](https://github.com/yangxuilyx/Desktop.Ava)

### DIFF
--- a/CefGlue.BrowserProcess/CefGlue.BrowserProcess.csproj
+++ b/CefGlue.BrowserProcess/CefGlue.BrowserProcess.csproj
@@ -1,66 +1,67 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
-  <PropertyGroup>
-    <OutputType>WinExe</OutputType>
-    <TargetFramework>$(DotnetVersion)</TargetFramework>
-    <AssemblyName>Xilium.CefGlue.BrowserProcess</AssemblyName>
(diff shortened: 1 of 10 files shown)


joaompneves:
There are some issues with that approach, which we considered in the past but abandoned, due to:
1) the BrowserProcess need to be built with the exact same version of the main app, otherwise we had issues, because of versions mismatch.
2) in mac the story is different and more complex, the folder structure is more restrict, and we hard time to get it working. Btw tried your branch and doesn't even build on mac.

yangxuilyx:
Maybe we can try to solve these issues:
1. BrowserProcess tried to set RollForward:major to allow the main app to be higher than its version, are there any other issues ?
2. sorry it didn't pass on mac, could you give me some advice on trying to fix the error on mac?

PR #208, opened 2025-04-02
RDEV-8412 - Bump to CEF v134.3.9

helderjgoncalves:
Description
--------
**GPU issues from CEF M121)** ✔️ 

Found a solution [here](https://github.com/cefsharp/CefSharp/discussions/4707#discussioncomment-12869815).

--------
**macOS vulkan issues** ✔️ 

This issue seems related with that Avalonia Sample App. 
The fix was to start to use Avalonia `ReactiveUI` framework.
```
/Users/hgo/CefGlue/CefGlue.Demo.Avalonia/bin/ARM64/Debug_WindowlessRender/net8.0/Xilium.CefGlue.Demo.Avalonia
objc[39249]: Class ExtensionDropdownHandler is implemented in both /Users/hgo/CefGlue/CefGlue.Demo.Avalonia/bin/ARM64/Debug_WindowlessRender/net8.0/runtimes/osx/native/libAvaloniaNative.dylib (0x1237760e0) and /Users/hgo/CefGlue/CefGlue.Demo.Avalonia/bin/ARM64/Debug_WindowlessRender/net8.0/libcef.dylib (0x33f1ce3b0). One of the two will be used. Which one is undefined.
2025-04-15 20:18:33.884 Xilium.CefGlue.Demo.Avalonia[39249:997598] +[IMKClient subclass]: chose IMKClient_Modern
2025-04-15 20:18:33.884 Xilium.CefGlue.Demo.Avalonia[39249:997598] +[IMKInputSession subclass]: chose IMKInputSession_Modern
[39304:259:0415/201834.846656:ERROR:angle_platform_impl.cc(49)] vk_renderer.cpp:184 (VerifyExtensionsPresent): Extension not supported: VK_EXT_metal_surface
[39304:259:0415/201834.846891:ERROR:angle_platform_impl.cc(49)] vk_renderer.cpp:184 (VerifyExtensionsPresent): Extension not supported: VK_KHR_surface
[39304:259:0415/201834.846918:ERROR:angle_platform_impl.cc(49)] Display.cpp:1085 (initialize): ANGLE Display::initialize error 0: Internal Vulkan error (-7): A requested extension is not supported, in ../../third_party/angle/src/libANGLE/renderer/vulkan/vk_renderer.cpp, enableInstanceExtensions:2144.
[...]

PR #226, opened 2025-09-15
Bump to CEF v139.0.28

trivalik:
I improved the cef 134 version from https://github.com/OutSystems/CefGlue/pull/208
with credits to: heldergoncalves92 filipnavara

mac cef package not tested.


helderjgoncalves:
@trivalik Thank you for the work invested in this CEF bump!
I'll be revisiting this topic shortly. And the goal is to make the CEF bump official.

I'll for sure review the work developed here.
Thank you!

PR #252, opened 2026-09-08
Fix high idle CPU from 1 kHz CEF message pump on macOS

paulomf:
On macOS (`ExternalMessagePump = true`), `AvaloniaBrowserProcessHandler` used the delay CEF
requests as the **period** of a recurring `Observable.Interval`. That delay is a deadline, not
an interval, and CEF requests `0` for the vast majority of calls, so the `delayMs <= 0 -> 1`
floor installed a permanent **1 kHz** pump. An idle app sat at roughly half a CPU core, with
every tick allocating on a thread-pool worker and posting to the UI thread.

Reusing the value as a period is wrong in both directions: short values spin the CPU, and CEF
also requests delays of several seconds, which as a period would stall the UI.

This polls at a fixed 30 Hz instead, matching the reference implementations — cefclient's
`MainMessageLoopExternalPump` (`kMaxTimerDelay = 1000 / 30`) and CefSharp's
`WpfBrowserProcessHandler` — and additionally pumps immediately when CEF asks for work now.
The timer is created once rather than being torn down and rebuilt on every notification.

The interval can be overridden through `AppContext` or the `CEFGLUE_MESSAGE_PUMP_INTERVAL_MS`
environment variable for hosts that want 60 Hz or for diagnostics, clamped to 16-1000 ms so
configuration cannot reintroduce either failure mode.

Measured on a host application built against this change, idling with a page loaded:

| | CEF subprocesses | idle CPU (of one core) |
| --- | --- | --- |
| before | 7 | 75.7% |
| after | 6 | 1.6% |

Interactivity was verified rather than assumed: clicking into a page input and typing still
filters and repaints the content.

Windows and Linux are unaffected, since they use `MultiThreadedMessageLoop` and never call
`OnScheduleMessagePumpWork`.


OS-rubencerqueira (dismissed):

helderjgoncalves (dismissed):

helderjgoncalves:
Reviewed this closely given the blast radius, and pushed 218aeea onto the branch with two changes to the immediate-pump path. Revert it if you would rather handle these yourself.

Verified the containment first, and it holds: `AvaloniaBrowserProcessHandler` is instantiated only in the `AvaloniaCefBrowser` static constructor behind a `CefRuntime.Platform == MacOS` check, so there is exactly one handler and one pump per process, and Windows/Linux never reach `OnScheduleMessagePumpWork` at all.

### `DispatcherPriority.Normal` is not Avalonia's normal priority

Avalonia does not reuse WPF's numbering. From `Avalonia.Base/Threading/DispatcherPriority.cs` (11.0.9):

| Priority | Value | Dispatched |
| --- | --- | --- |
| `Send` | 6 | unconditionally |
| `Normal` | 5 | unconditionally |
| `Render` | 4 | unconditionally |
| `Loaded` | 1 | unconditionally |
| `Default` | 0 | unconditionally |
| `Input` | -1 | only when no pending OS input |
| `Background` | -2 | only when no pending OS input |

`Default` is the normal priority here; `Normal` is a WPF-compatibility alias sitting second from the top, above `Render`. So the immediate post was letting CEF preempt layout and compositing, and it was doing so on the path that fires most often — as the description notes, CEF asks for a `0` delay the vast majority of the time. Under sustained page activity the pump re-triggers `OnScheduleMessagePumpWork(0)` from inside `DoMessageLoopWork`, so priority-5 jobs can keep getting picked ahead of pending `Render` work.

`Background` looks like the obvious correction, since `AvaloniaScheduler` already puts the 30 Hz timer there — but it is wrong for this path, and `Input` is not a middle ground. `Dispatcher.ExecuteJobsCore` splits the queue at `Input`, not at `Background`:

```csharp
// We don't stop for executing jobs queued with >Input priority
if (job.Priority > DispatcherPriority.Input)
{
    ExecuteJob(job);
}
// If platform supports pending input query, ask the platform if we can continue running low priority jobs
else if (_pendingInputImpl?.CanQueryPendingInput == true)
{
    if (!_pendingInputImpl.HasPendingInput)
        ExecuteJob(job);
    else
    {
        RequestBackgroundProcessing();
        return;
    }
}
```

Anything at `Input` or below is held back while the platform reports pending OS input, which would stall the pump precisely while the user is typing or scrolling. `Default` is the lowest priority still dispatched unconditionally, and it sits below `Loaded`, `UiThreadRender`, `AfterRender` and `Render`, so it cannot reintroduce the jank that `Normal` risked.

### The immediate path had no coalescing

Before this PR that was implicit — each notification disposed the previous subscription, so at most one pump was pending. Creating the timer once is the right call, but the new code posts an unconditional dispatcher job per notification, and `OnScheduleMessagePumpWork` is documented as "called from any thread". A burst therefore queues N independent `DoMessageLoopWork` jobs even though the first drains everything pending. cefclient guards this with its `is_active_` / `reentrancy_detected_` pair; an `Interlocked` flag is the cheap equivalent, cleared *before* the pump runs so work CEF schedules from inside it still queues a fresh notification rather than being dropped.

### Left alone, your call

- The 30 Hz fallback timer stays at `Background`, because that is what `AvaloniaScheduler` hardcodes and changing it means dropping the Rx interval. Fine while it is only a backstop, but it does mean the `delayMs > 0` case is input-gated. A `DispatcherTimer` at `Default` would close that and also drop a thread-pool wakeup plus a cross-thread post per tick.
- `_pump` is never disposed, so it keeps ticking after `CefRuntime.Shutdown()`, and `DoMessageLoopWork` has no initialization guard. Not a regression — the old code leaked its last subscription too — and the only shutdown path is the `ProcessExit` handler in `CefRuntimeLoader`, by which point the dispatcher no longer delivers posts. Low risk, but this PR is the natural place to fix it.
- `ResolvePumpIntervalMs` clamps silently, so setting `5` yields `16` with no diagnostic, and an unparsable `AppContext` value never falls through to the environment variable.
- The version goes `.211` to `.213`, skipping `.212` — deliberate?

Builds clean. @OS-rubencerqueira, sorry, this will have staled your approval.

helderjgoncalves:
Worked through what actually flows down each of the two paths, since the split between them is the part with the least obvious consequences. Three concrete scenarios, plus two clarifications on the timer.

### 1. JS calling back into C# — immediate path, slightly better than before

A React handler invokes a native object method. That is serialized into a `CefProcessMessage` in the render subprocess, arrives on CEF's IO thread, and is posted to the browser UI thread task runner. It is runnable now, so CEF calls `OnScheduleMessagePumpWork(0)` — from that IO thread, which is precisely why the notification has to be marshalled rather than pumped inline.

We claim the coalescing slot, post at `Default`, and the pump runs on the next dispatcher turn. Previously this waited on the 1 ms timer, so it is marginally quicker now and without the 1 kHz spin underneath. Same shape for `EvaluateScriptFunction` responses, custom scheme handlers, and paint callbacks.

### 2. Debounced input, caret blink, keep-alives — next tick, and it does not matter

A React `setTimeout(..., 300)` puts a task in Blink's timer queue and CEF calls `OnScheduleMessagePumpWork(300)`. We take the early return and post nothing; the 30 Hz timer pumps and the work runs at the first tick at or after the deadline, so 300 ms becomes 300-333 ms. Imperceptible, and this is the overwhelmingly common shape of delayed work.

### 3. Timer-paced animation — next tick, and this one is worth a look

:warning: This is the one I would flag as a possible consequence of the change.

A CSS animation or `requestAnimationFrame` loop wants the next frame in roughly 16 ms for 60 fps, so CEF asks for ~16 ms. We round that up to the next 30 Hz tick, so the animation is paced at ~33 ms — 30 fps rather than 60.

This is where we diverge from cefclient:

```cpp
if (delay_ms <= 0) {
  // Execute the work immediately.
  DoWork();
} else {
  // Never wait longer than the maximum allowed time.
  if (delay_ms > kMaxTimerDelay) {
    delay_ms = kMaxTimerDelay;
  }
  SetTimer(delay_ms);
}
```

`kMaxTimerDelay` (`1000 / 30`) is a **cap on a one-shot timer**, not a period. cefclient serves a 16 ms request at 16 ms and only clamps the long ones; we serve everything at a fixed 33 ms granularity. `main` was also accurate on the first fire here — its bug was the repetition, not the initial delay — so this specific aspect is a small regression against both.

Two things soften it. A good deal of compositor work arrives as `delay 0` and goes down the immediate path untouched, so it is specifically *timer-paced* animation that halves, not everything that moves. And the verification on this PR (clicking into an input and typing) exercises scenario 1, not scenario 3, so it would not have surfaced either way.

If it is worth closing, the cefclient shape is cheap: on `delayMs > 0`, arm a one-shot at `min(delayMs, PumpIntervalMs)` and keep the free-running timer underneath.

### Two clarifications on the timer

It is not a fallback or a backstop: it is the sole mechanism serving every `delayMs > 0` notification, since that path posts nothing at all. That matters for the `Background` priority question, because the input-gating then applies to all delayed work rather than only to a safety net.

And the free-running design matches cefclient rather than departing from it. `DoWork()` ends with `else if (!IsTimerPending()) OnScheduleMessagePumpWork(kTimerDelayPlaceholder)`, where the placeholder is `INT_MAX` and gets clamped to 33 ms — so cefclient re-arms after every pump and never goes quiet either.

helderjgoncalves (dismissed):

helderjgoncalves (approved):

PR #254, opened 2026-09-21
Pump CEF work at Default priority and coalesce immediate pumps

helderjgoncalves:
Follow-up to #252, stacked on its branch because it refines code that only exists there. Review this one against `fix/macos-message-pump-idle-cpu`, not `main`.

#252 is the right fix for the 1 kHz pump. Two things in the new immediate-pump path are worth tightening before it ships.

## `DispatcherPriority.Normal` is not Avalonia's normal priority

Avalonia does not reuse WPF's numbering. From `Avalonia.Base/Threading/DispatcherPriority.cs` (11.0.9):

| Priority | Value | Dispatched |
| --- | --- | --- |
| `Send` | 6 | unconditionally |
| `Normal` | 5 | unconditionally |
| `Render` | 4 | unconditionally |
| `Loaded` | 1 | unconditionally |
| `Default` | 0 | unconditionally |
| `Input` | -1 | only when no pending OS input |
| `Background` | -2 | only when no pending OS input |

`Default` is the "normal" priority in Avalonia; `Normal` is a WPF-compatibility alias sitting second from the top, above `Render`.

So `Post(CefRuntime.DoMessageLoopWork, DispatcherPriority.Normal)` lets CEF preempt layout and compositing — and it does so on the path that fires most often, since as #252 notes CEF asks for a `0` delay the vast majority of the time. Under sustained page activity the pump re-triggers `OnScheduleMessagePumpWork(0)` from inside `DoMessageLoopWork`, so priority-5 jobs can keep getting picked ahead of pending `Render` work.

## Why `Default` and not `Background` or `Input`

`Background` looks like the natural choice, since the 30 Hz timer in the same method already runs there — `AvaloniaScheduler` hardcodes `DispatcherPriority.Background`. It is the wrong choice for the immediate path, and `Input` is not a middle ground. `Dispatcher.ExecuteJobsCore` splits the queue at `Input`, not at `Background`:

```csharp
// We don't stop for executing jobs queued with >Input priority
if (job.Priority > DispatcherPriority.Input)
{
    ExecuteJob(job);
}
// If platform supports pending input query, ask the platform if we can continue running low priority jobs
else if (_pendingInputImpl?.CanQueryPendingInput == true)
{
    if (!_pendingInputImpl.HasPendingInput)
        ExecuteJob(job);
    else
    {
        RequestBackgroundProcessing();
        return;
    }
}
```

Anything at `Input` (-1) or below is held back while the platform reports pending OS input, and `RequestProcessing` routes it through `RequestBackgroundProcessing` instead of signalling the loop. That would defer the message pump precisely while the user is typing or scrolling — the worst possible moment for a browser.

`Default` (0) is the lowest priority that is still dispatched unconditionally, and it sits below `Loaded`, `UiThreadRender`, `AfterRender` and `Render`, so it cannot reintroduce the jank that `Normal` risked.

## The immediate path had no coalescing

Before #252 this was implicit: each notification disposed the previous subscription, so at most one pump was ever pending. #252 creates the timer once (good) but posts an unconditional dispatcher job per notification. `OnScheduleMessagePumpWork` is documented as "called from any thread", so a burst queues N independent `DoMessageLoopWork` jobs even though the first one drains everything pending.

cefclient guards this with its `is_active_` / `reentrancy_detected_` pair. An `Interlocked` flag is the cheap equivalent: one pump in flight at a time, flag cleared *before* `DoMessageLoopWork` runs so work CEF schedules from inside the pump still queues a fresh notification rather than being dropped.

## Scope

macOS only, same as #252 — `AvaloniaBrowserProcessHandler` is instantiated solely in the `AvaloniaCefBrowser` static constructor behind a `CefRuntime.Platform == MacOS` check, and `ExternalMessagePump` is only set on macOS.

Builds clean. No version bump: this is meant to land with #252 under its `120.6099.213`.

## Open question for @paulomf

The 30 Hz fallback timer stays at `Background`, because that is what `AvaloniaScheduler` gives you and changing it means dropping the Rx interval. That is fine while it is only a backstop, but it does mean the `delayMs > 0` case — where we rely purely on the timer — is input-gated. If that matters, the cleaner shape is a `DispatcherTimer` at `Default` instead of `Observable.Interval(...).ObserveOn(AvaloniaScheduler.Instance)`, which would also drop a thread-pool wakeup and a cross-thread post per tick. Happy to fold that in.

## Not addressed here

- `_pump` is never disposed, so it keeps ticking after `CefRuntime.Shutdown()` and `DoMessageLoopWork` has no initialization guard. Not a regression — the old code leaked its last subscription too — and the only shutdown path is the `ProcessExit` handler in `CefRuntimeLoader`, by which point the dispatcher no longer delivers posts.
- `ResolvePumpIntervalMs` silently clamps, so setting `5` yields `16` with no diagnostic, and an unparsable `AppContext` value never falls through to the environment variable.


helderjgoncalves:
Superseded: folding these two changes directly into #252 rather than stacking a second PR on top of it.

